PR #1, opened 2019-03-05
CDAP-14932 Salesforce Batch Source to fetch objects via SOQL. Initial structure.

aonischuk:
Build related stuff. 

In pom.xml I added some dependencies which will be needed in future/already used in my draft code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,32 @@
+# intellij
+.idea/
+target/*
+force/target/*
+*.iml
+
+# Compiled class file
+*.class
+
+# Log file
+*.log
+
+# BlueJ files
+*.ctxt
+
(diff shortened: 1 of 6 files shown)


albertshau:
That header was copy-pasted from somewhere else, please keep using Cask Data. If we switch over we'll do them all at the same time.

aonischuk:
@albertshau fixed the headers.

aonischuk:
Also I saw that package name Bhooshan had was com.google.plugins not co.cask.plugins. Should I use cask as well?

albertshau:
yep

aonischuk:
@albertshau fixed that as well. Does the patch look good to go?

PR #2, opened 2019-03-07
CDAP-14932 Salesforce Batch Source to fetch objects via SOQL. Authentication via oauth2

aonischuk:
Authentication via oauth2 to salesforce

### DIFF
--- a/src/main/java/co/cask/hydrator/salesforce/authenticator/Authenticator.java
+++ b/src/main/java/co/cask/hydrator/salesforce/authenticator/Authenticator.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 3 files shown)


src/main/java/co/cask/hydrator/salesforce/authenticator/Authenticator.java
```diff
@@ -0,0 +1,82 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.salesforce.authenticator;
+
+import co.cask.hydrator.salesforce.SalesforceUtils;
+import com.google.gson.Gson;
+import com.sforce.ws.ConnectorConfig;
+import org.eclipse.jetty.client.HttpClient;
```
jayfjin: eclipse? is this the right class?

aonischuk: in most examples salesforce will use jetty as a server. Which is developed by eclipse.
That's why I used it. Is this problematic?

PR #13, opened 2019-04-15
CDAP-15247 Remove spark-plugins:jar:2.2.0-SNAPSHOT from dependencies

aonischuk:
This dependency is broken in repo. So unless installed locally the build fails: 

[WARNING] The POM for co.cask.hydrator:spark-plugins:jar:2.2.0-SNAPSHOT is invalid, transitive dependencies (if any) will not be available: 1 problem was encountered while building the effective model for co.cask.hydrator:spark-plugins:2.2.0-SNAPSHOT
...
[ERROR] Failed to execute goal on project salesforce-plugins: Could not resolve dependencies for project co.cask.hydrator:salesforce-plugins:jar:1.0.0-SNAPSHOT: Could not find artifact co.cask.hydrator:spark-plugins:jar:2.2.0-SNAPSHOT -> [Help 1]
[ERROR]

For other streaming plugin seems like this dependency is not used as well. And we have only small class we use from it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,11 +190,6 @@
       <version>${spark2.version}</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>co.cask.hydrator</groupId>
-      <artifactId>spark-plugins</artifactId>
-      <version>${hydrator.version}</version>
-    </dependency>
 
     <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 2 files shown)


jayfjin:
FWIW we need to rename the dependencies to io.cdap from co.cask.cdap as detailed in Maksym's email thread. However, does this change work if we remove this dependency? (Build, test usage, etc)

aonischuk:
yes dependency contained only one class. Usage of which I removed.
Also I have tested this to work in pipeline + ran unit tests.

jayfjin:
Ok this is fine to do here. For other cases where we actually depend on the package we should be renaming it though.

PR #15, opened 2019-04-20
package renaming

anwar6953:


aielchiieva:
@anwar6953 / @jayfjin I think we should also merge these changes into master. For example, in the scope of multi source support task, I am going to add new classes and I think its better if they already use new package naming.

PR #16, opened 2019-04-23
CDAP-15030. Salesforce Streaming Source to fetch data. Use logical types/Handle unexpected json fields

aonischuk:
Bugfixes:
- use logical types for time and date
- handle Opportunity.Fiscal field. This field has format declared as "string", but is returned via Streaming API as json object. Looks like some kind of undocumented behavior from Salesforce. Which needs to be handled in pipeline in runtime.
- log previously ignored exceptions from salesforce spark receiver thread

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceReceiver.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceReceiver.java
@@ -72,8 +72,11 @@ public class SalesforceReceiver extends Receiver<String> {
           store(message);
         }
       }
-    } catch (InterruptedException e) {
-      throw new RuntimeException("Exception while receiving messages from pushTopic", e);
+    } catch (Exception e) {
+      String errorMessage = "Exception while receiving messages from pushTopic";
+      // Since it's top level method of thread, we need to log the exception or it will be unseen
+      LOG.error(errorMessage, e);
+      throw new RuntimeException(errorMessage, e);
     }
   }
 }
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
@@ -45,8 +45,12 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Instant;
+import java.time.LocalTime;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 import javax.ws.rs.Path;
 
 /**
@@ -137,7 +141,7 @@ public class SalesforceStreamingSource extends StreamingSource<StructuredRecord>
           continue; // this field is not in schema
         }
 
-        builder.set(field.getName(), value);
+        builder.set(field.getName(), convertValue(value, field));
       }
       return builder.build();
     } catch (Exception ex) {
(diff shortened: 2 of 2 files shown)


src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
```diff
@@ -154,6 +158,51 @@ private StructuredRecord getStructuredRecord(String jsonMessage) throws Connecti
...
+
+    if (fieldSchema.getLogicalType() != null) {
+      String valueString = (String) value;
+      switch (logicalType) {
+        case DATE:
+          return Math.toIntExact(ChronoUnit.DAYS.between(Instant.EPOCH, Instant.parse(valueString)));
+        case TIMESTAMP_MICROS:
+          return TimeUnit.MILLISECONDS.toMicros(Instant.parse(valueString).toEpochMilli());
+        case TIME_MICROS:
+          return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(valueString).toNanoOfDay());
```
jayfjin: Is toNano consistent with other conversions of TIME_MICROS?

src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
```diff
@@ -154,6 +158,51 @@ private StructuredRecord getStructuredRecord(String jsonMessage) throws Connecti
...
+    }
+
+    Schema.Type fieldSchemaType = fieldSchema.getType();
+    Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+
+    if (fieldSchema.getLogicalType() != null) {
+      String valueString = (String) value;
+      switch (logicalType) {
+        case DATE:
+          return Math.toIntExact(ChronoUnit.DAYS.between(Instant.EPOCH, Instant.parse(valueString)));
```
jayfjin: I see a lot of different time classes being used. Is it possible to use a class already used for this instead of bringing in a new one?

src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
```diff
@@ -154,6 +158,51 @@ private StructuredRecord getStructuredRecord(String jsonMessage) throws Connecti
...
+          return TimeUnit.MILLISECONDS.toMicros(Instant.parse(valueString).toEpochMilli());
+        case TIME_MICROS:
+          return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(valueString).toNanoOfDay());
+        default:
+          throw new UnexpectedFormatException(String.format("Field '%s' is of unsupported type '%s'",
+                                                            field.getName(), logicalType.getToken()));
+      }
+    }
+
+    // Found a single field (Opportunity.Fiscal) which is documented as string and has a string type
```
jayfjin: Let's also document this in the readme

src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
aonischuk: it converts nanos to micros the method used is "toMicros". So it is consistent.

src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
aonischuk: Basically here he we are calculating how many days since 1970 passed to the given date. I find one method which does not require that other lib. But it requires timezone which we don't really want, since we don't know which timezone is that time from.

src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforceStreamingSource.java
aonischuk: Do you think we should add this to a top level readme? This is a very specific small salesforce issue which is only specific to this plugin (not others) because of using streaming API. Not sure where to put it and where even such a small specific info should go into.

jayfjin: I see - if it's only for streaming it's fine. Otherwise we should add it to the types map in the readme

PR #20, opened 2019-04-26
CDAP-15130: Implement Salesforce Batch Multi Source plugin

aielchiieva:
1. Factored out common logic for Salesforce Batch Source and Multi Source plugins into base package.
2. Added Salesforce Batch Multi Source classes into multi package.
3. Added integration tests.
4. Fixed parser to cover cases when SObject contains field with the same name as SObject name and when SObject contains field named Category which was treated as keyword in parser.


docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -0,0 +1,121 @@
...
+    | 0      | shirt123 | view   |
+    | 0      | carxyz   | view   |
+    | 0      | shirt123 | buy    |
+    | 0      | coffee   | view   |
+    | 1      | cola     | buy    |
+    +--------+----------+--------+
+    
+To read data from these two SObjects, both of them must be indicated in the `White List` configuration property.
+
+The output of the the source will be the following records:
```
jayfjin: I think it will be helpful to also show here how the schemas are merged into the record (with the UUID of each).

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseBatchSource.java
```diff
@@ -0,0 +1,148 @@
...
+
+    // empty string is considered null in csv
+    if (Strings.isNullOrEmpty(value)) {
+      return null;
+    }
+
+    Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+    if (fieldSchema.getLogicalType() != null) {
+      switch (logicalType) {
+        case DATE:
```
jayfjin: I've seen this switch case structure around a few times now in various code reviews. Does this refactor capture all usages of it? If not we should extract this to a util/helper class

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseSourceConfig.java
```diff
@@ -0,0 +1,145 @@
...
+   * Generates SOQL based on given sObject name metadata and filter properties.
+   * Includes only those sObject fields which are present in the schema.
+   * This allows to avoid pulling data from Salesforce for the fields which are not needed.
+   *
+   * @param sObjectName Salesforce object name
+   * @param schema CDAP schema
+   *
+   * @return SOQL generated based on sObject metadata and given filters
+   */
+
```
jayfjin: nit: extra newline

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/multi/SalesforceMultiInputFormatProvider.java
```diff
@@ -0,0 +1,95 @@
...
+    Schema fieldSchema = field.getSchema();
+    if (fieldSchema.getRecordName() == null || fieldSchema.getFields() == null) {
+      throw new IllegalArgumentException(
+        String.format("Field '%s' if of invalid schema '%s'", field.getName(), fieldSchema));
+    }
+    List<Schema.Field> fields = new ArrayList<>(fieldSchema.getFields());
+    fields.add(Schema.Field.of(sObjectNameField, Schema.of(Schema.Type.STRING)));
+    return Schema.Field.of(field.getName(), Schema.recordOf(fieldSchema.getRecordName(), fields));
+  }
+
```
jayfjin: nit: extra newline

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchMultiSourceETLTest.java
```diff
@@ -0,0 +1,137 @@
...
+        .build());
+
+    addSObjects(sObjects, false);
+
+    List<StructuredRecord> records = getResultsForMultiSObjects(
+      String.join(",", sObjectName1, sObjectName2), null);
+
+    String sObjectNameField = SalesforceMultiSourceConfig.SOBJECT_NAME_FIELD_DEFAULT;
+
+    Set<String> expectedResults = Sets.newHashSet(
```
jayfjin: Instead of converting the record into a string, can we build what the expected records should look like for each, and then compare the records?

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
```diff
@@ -79,7 +81,38 @@ public static Schema getSchema(AuthenticatorCredentials credentials, SObjectDesc
     SObjectsDescribeResult describeResult = new SObjectsDescribeResult(partnerConnection,
       sObjectDescriptor.getAllParentObjects());
 
-    return getSchemaWithFields(sObjectDescriptor, describeResult);
+    return getSchemaWithFields(sObjectDescriptor, describeResult, "output");
+  }
+
+  /**
+   * Generates CDAP schema for multiple SObjects, where each field of type record represents one SObjects.
```
jayfjin: Is this only used to generate CDAP schema in the multiple SObject case? If so imo the function name should reflect that

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseBatchSource.java
```diff
@@ -0,0 +1,148 @@
...
+      case INT:
+        return Integer.parseInt(value);
+      case LONG:
+        return Long.parseLong(value);
+      case FLOAT:
+        return Float.parseFloat(value);
+      case DOUBLE:
+        return Double.parseDouble(value);
+      case BYTES:
+        return Byte.parseByte(value);
```
albertshau: this is wrong, it is parsing a single byte whereas 'bytes' is a byte array. Does salesforce have a bytes type?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseBatchSource.java
```diff
@@ -0,0 +1,148 @@
...
+import java.time.LocalTime;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Base class for Salesforce Batch Source plugins. Contains data transformation logic.
+ *
+ * @param <T> input data key value
+ */
+public abstract class SalesforceBaseBatchSource<T> extends BatchSource<T, Map<String, String>, StructuredRecord> {
```
albertshau: Instead of an abstract base plugin class, this logic should be pulled out into a class that transforms a single line from the bulk API response to a StructuredRecord.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseSourceConfig.java
```diff
@@ -0,0 +1,145 @@
...
+      List<String> fieldNames;
+      if (schema == null) {
+        fieldNames = sObjectFields;
+      } else {
+        fieldNames = sObjectFields.stream()
+          .filter(name -> schema.getField(name) != null)
+          .collect(Collectors.toList());
+
+        if (fieldNames.isEmpty()) {
```
albertshau: Is there validation done somewhere checking if the schema has fields that are not present in the SObject?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/base/SalesforceBaseSourceConfig.java
```diff
@@ -0,0 +1,145 @@
...
+              + " Schema: '%s'. SObject fields: '%s'", schema, sObjectFields));
+        }
+      }
+
+      String sObjectQuery = SalesforceQueryUtil.createSObjectQuery(fieldNames, sObjectName,
+        getDuration(), getOffset(), getDatetimeFilter());
+      LOG.debug("Generated SObject query: '{}'", sObjectQuery);
+      return sObjectQuery;
+    } catch (ConnectionException e) {
+      throw new IllegalStateException(
```
albertshau: doesn't look like this is adding much value, this method should just throw the ConnectionException.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/multi/SalesforceMultiInputFormat.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class SalesforceMultiInputFormat extends SalesforceBaseInputFormat {
+
+  private static final Gson GSON = new Gson();
+  private static final Type QUERIES_TYPE = new TypeToken<List<String>>() { }.getType();
+
+  @Override
+  public List<InputSplit> getSplits(JobContext context) {
+    Configuration configuration = context.getConfiguration();
+    List<String> queries = GSON.fromJson(configuration.get(SalesforceSourceConstants.CONFIG_QUERIES), QUERIES_TYPE);
```
albertshau: Instead of having separate Multi classes for the input format, record reader, split, and inputformat provider, it is much simpler to just add the query to the SalesforceSplit and enhance the existing inputformat to support multiple queries. The multi source would just configure it with multiple queries while the single source would configure it with a single query.

aielchiieva: Done, though I felt code was less coupled with previous design.

aielchiieva:
@albertshau 

> On a separate note, I looked over the existing SalesforceRecordReader and noticed it reads the entire Bulk API response into a single string, then splits and parses.
> 
> This is not correct, the record reader should be reading the response in a streaming fashion, with each call to nextKeyValue reading the next line of input from the input stream, and close closing the input stream. Please open a jira and fix it in a separate PR.

It works not quite as you have described. `SalesforceInputFormat#getSplits` submits Bulk API query and receives batchId array. For each batchId new input split is created. `SalesforceRecordReader` queries Salesforce using batchId from input split. As result it receives list of input streams for one batchId. Since all input streams are the parts of one CSV file they should be read into a single string to be parsed properly using CSV reader.

aielchiieva:
@albertshau / @jayfjin addressed code review comment, please review.

docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -0,0 +1,121 @@
...
+
+**Duration:** SObject query duration filter is applied to system field `LastModifiedDate` to allow range query. 
+Duration units set to `hours`. For example, if duration is '6' (6 hours) and the pipeline runs at 9am, it will read data 
+updated from 3am - 9am. Ignored if `datetimeFilter` is provided.
+
+**Offset:** SObject query offset filter is applied to system field `LastModifiedDate` to allow range query. 
+Offset units set to `hours`. For example, if duration is '6' (6 hours) and the offset is '1' (1 hour) and the pipeline 
+runs at 9am, it will read data updated from 2am - 8am. Ignored if `datetimeFilter` is provided.
+
+**Error Handling:** Strategy used to handle erroneous records. Acceptable values are Skip on error,
```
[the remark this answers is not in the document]
aonischuk: This function was requested by Nitin Motgi 
https://github.com/data-integrations/salesforce/pull/4#discussion_r268378379

The error happens if the csv row is invalid. For example not enough values or too much values or casting to type from schema is not valid (for example expected in found string)

docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -0,0 +1,121 @@
...
+To create Client Secret user needs to create a connected application in Salesforce first.
+
+**Login Url:** Salesforce OAuth2 login url.
+
+**White List**: List of SObjects to replicate from Salesforce. By default all SObjects will be white listed. 
+For each white listed SObject, SOQL query will be generated: `select <FIELD_1, FIELD_2, ..., FIELD_N> from ${sObjectName}`.
+
+**Black List**: List of SObjects NOT to replicate from Salesforce. By default NONE of SObjects will be black listed.
+
+**Datetime Filter:** SObject query datetime filter is applied to system field `LastModifiedDate` to allow incremental
```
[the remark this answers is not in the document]
albertshau: OK let's open a jira and fix separately.

After reading the doc again I understand better, this is only used when an SObject is given and not for SOQL. I think the confusing part is that users can either specify a DatetimeFilter (which is really just specifying the start time), or they can specify a duration and offset.

I think the confusion can be removed if instead of these three, we have a 'start' and 'end' property. If 'start' is not specified, it reads from the start of time. If 'end' is not specified, it reads up until the logical start time of the pipeline (context.getLogicalStartTime()). This way, 'start' replaces the current DatetimeFilter, and it is probably easier to understand than duration and offset.

docs/SalesforceMultiObjects-batchsource.md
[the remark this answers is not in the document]
albertshau: but doesn't the schema come from salesforce? Does it somehow allow its data to not match its schema?

albertshau:
At https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceRecordReader.java#L78, it is reading the entire contents of the response into memory, then parsing.

https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java#L113 is a bad method and should not exist. The InputStream from https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java#L127 should be used to read the next line whenever the record reader gets the next key value, and closed when the record reader is closed.

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchMultiSourceETLTest.java
albertshau: If the values are just strings, you be able to build the records like Jay mentioned and Assert.assertEquals(expectedRecords, actualRecords).

It's only if the records contain bytes, arrays, and other things that don't compare nicely in Java that you need to explicitly compare field values.

albertshau:
just one comment regarding one of Jay's comments in the test

aielchiieva:
@albertshau updated integration test. If all changes look good, I will squash the commits and resolve the conflicts.

aielchiieva:
@albertshau 
> At https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceRecordReader.java#L78, it is reading the entire contents of the response into memory, then parsing later in the record reader.
> 
> https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java#L113 is a bad method and should be deleted. The InputStream from https://github.com/data-integrations/salesforce/blob/master/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java#L127 should be used to read the next line whenever the record reader gets the next key value, and closed when the record reader is closed.

I have created Jira for further discussion and fix - https://issues.cask.co/browse/CDAP-15319.

docs/SalesforceMultiObjects-batchsource.md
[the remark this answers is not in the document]
aonischuk: Yes schema comes from Salesforce. I don't think it should ever happen in normal flow.

We just had some handling for unexpected values in CSV and Nitin asked to implement error handling like that. 

@nitinmotgi do you think we can remove errorHandling due to reasons Albert mentions or you have some other cases in mind?

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchMultiSourceETLTest.java
aielchiieva: Done.

docs/SalesforceMultiObjects-batchsource.md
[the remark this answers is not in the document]
aielchiieva: Created Jira for the fix: https://issues.cask.co/browse/CDAP-15320

docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -0,0 +1,121 @@
...
+    |                                  | YYYY-MM-DDThh:mm:ss-hh:mm | 1999-01-01T23:01:01-08:00 |
+    |                                  | YYYY-MM-DDThh:mm:ssZ      | 1999-01-01T23:01:01Z      |
+    +----------------------------------+---------------------------+---------------------------+
+
+`SOQL Date Literal` - fieldExpression to compare a range of values to the value in a datetime 
+field. Each literal is a range of time beginning with midnight (00:00:00). For example: `YESTERDAY`, `LAST_WEEK`, 
+`LAST_MONTH` ...
+
+**Duration:** SObject query duration filter is applied to system field `LastModifiedDate` to allow range query. 
+Duration units set to `hours`. For example, if duration is '6' (6 hours) and the pipeline runs at 9am, it will read data
```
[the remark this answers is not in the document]
aielchiieva: Created Jira for the fix: https://issues.cask.co/browse/CDAP-15320

aielchiieva:
@albertshau / @jayfjin addressed remaining code review comments, for each mentioned enhancement has created Jiras (only one is under discussion about error handling, waiting for the comment from @nitinmotgi). Please let me know if there are any other changes required before I squash and resolve the conflicts.

docs/SalesforceMultiObjects-batchsource.md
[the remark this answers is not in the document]
albertshau: Let's open a jira for this so that it doesn't block this PR in the meantime.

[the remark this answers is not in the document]
aielchiieva: Created Jira - https://issues.cask.co/browse/CDAP-15325

aielchiieva:
@albertshau squashed commits and resolved the conflicts. PR is ready to be merged.

PR #22, opened 2019-05-01
CDAP-15321: Add sonatype repositories and update JUnit version

aielchiieva:


aielchiieva:
@albertshau / @jayfjin please review.

pom.xml
```diff
@@ -30,8 +30,8 @@
     <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
-    <cdap.version>6.0.0-SNAPSHOT</cdap.version>
-    <junit.version>4.11</junit.version>
+    <cdap.version>6.1.0-SNAPSHOT</cdap.version>
```
albertshau: I think it's better to keep it at 6.0.0-SNAPSHOT. Once CDAP 6.0.0 is released, it should be changed to 6.0.0

aielchiieva: @albertshau changed back version to `6.0.0-SNAPSHOT`.

PR #23, opened 2019-05-02
CDAP-15246. Fix macros checks, remove deprecated API usage, some minor fixes

aonischuk:
1. Fix Streaming plugin validation failure when macros are present in credentials.
2. When building Salesforce project, there is warning since the deprecated API is used.
`.../salesforce-plugins/src/main/java/co/cask/hydrator/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java uses or overrides a deprecated API.`
Rewrote the code to avoid using deprecated method org.cometd.bayeux.Message#getJson
For generating json from message in new API jackson or jetty has to be used. Using jackson in our case.
3. PushTopicNotifyCreate, pushTopicNotifyUpdate, pushTopicNotifyDelete are marked as nullable but based on widget configuration, they can not be null.
4. Rename constants "PUSHTOPIC" -> "PUSH_TOPIC"
```
  protected static final String PROPERTY_PUSHTOPIC_NAME = "pushTopicName";
  protected static final String PROPERTY_PUSHTOPIC_QUERY = "pushTopicQuery";
```
5. Add final:
`private static Pattern isValidFieldNamePattern = Pattern.compile("[a-zA-Z0-9.-_]+");`

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceConfig.java
@@ -51,39 +51,36 @@ public class SalesforceStreamingSourceConfig extends BaseSalesforceConfig implem
   private static final Logger LOG = LoggerFactory.getLogger(SalesforceStreamingSourceConfig.class);
   private static final long serialVersionUID = 4218063781902315444L;
 
-  protected static final String ENABLED_KEYWORD = "Enabled";
-  protected static final String PROPERTY_PUSHTOPIC_NAME = "pushTopicName";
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceConfig.java
```diff
@@ -275,6 +267,10 @@ private static void assertFieldValue(SObject pushTopic, String fieldName, Object
   }
 
   private String getSObjectQuery() {
```
albertshau: annotate this method as `@Nullable`

aonischuk: fixed

aonischuk:
@albertshau PR is ready to be merged.

PR #25, opened 2019-05-03
CDAP-15319 Salesforce Batch Source: read the response in a streaming fashion

aonischuk:
Previously for every batch we formed response as String and than read it.
The change makes RecordReader read directly from InputStream for a batch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
     <mockito.version>1.10.19</mockito.version>
     <commons.csv.version>1.6</commons.csv.version>
     <jackson.version>1.9.13</jackson.version>
-    <jackson2.version>2.9.8</jackson2.version>
+    <jackson2.version>2.6.7</jackson2.version>
     <json.version>20180813</json.version>
     <awaitility.version>3.1.6</awaitility.version>
     <commons-logging.version>1.2</commons-logging.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -44,7 +44,7 @@
     <mockito.version>1.10.19</mockito.version>
     <commons.csv.version>1.6</commons.csv.version>
     <jackson.version>1.9.13</jackson.version>
-    <jackson2.version>2.9.8</jackson2.version>
+    <jackson2.version>2.6.7</jackson2.version>
```
aonischuk: This is something I missed in previous PR. Unless this is done streaming integration tests fail with:

```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Incompatible Jackson version: 2.9.4
at com.fasterxml.jackson.module.scala.JacksonModule$class.setupModule(JacksonModule.scala:64)
at com.fasterxml.jackson.module.scala.DefaultScalaModule.setupModule(DefaultScalaModule.scala:19)
at com.fasterxml.jackson.databind.ObjectMapper.registerModule(ObjectMapper.java:751)
at org.apache.spark.rdd.RDDOperationScope$.(RDDOperationScope.scala:82)
```

While the production code is code is successful (which is what I tested when making changes)

aonischuk:
@albertshau can we this at this point?

PR #29, opened 2019-05-09
CDAP-15320: Enhanced SObject incremental and range filters

aielchiieva:
1. Replace 'Datetime filter' with 'Last Modified After' and 'Last Modified Before' filters
2. Drop support for the date literals (yesterday, last_month, etc)
3. Add multiple unit type support to duration and offset filters

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.salesforce;
+
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.Map;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * SObject query datetime filter handler. Can be of three types:
+ * <ul>
+ * <li>Interval - filter using provided start and end datetime</li>
+ * <li>Range - filter calculated from provided start time, duration and offset</li>
+ * <li>NoOp - filter initialized with null</li>
+ * </ul>
+ */
+public final class SObjectFilterDescriptor {
+
+  private static final SObjectFilterDescriptor NO_OP_FILTER_INSTANCE = new SObjectFilterDescriptor(null, null);
+
+  @Nullable
+  private final ZonedDateTime startTime;
+  @Nullable
+  private final ZonedDateTime endTime;
+
+  public static SObjectFilterDescriptor noOp() {
+    return NO_OP_FILTER_INSTANCE;
+  }
+
+  public static SObjectFilterDescriptor interval(@Nullable ZonedDateTime startTime,
+                                                 @Nullable ZonedDateTime endTime) {
+    return startTime == null && endTime == null
+      ? NO_OP_FILTER_INSTANCE
+      : new SObjectFilterDescriptor(startTime, endTime);
+  }
+
+  public static SObjectFilterDescriptor range(long logicalStartTime,
+                                              Map<ChronoUnit, Integer> duration,
+                                              Map<ChronoUnit, Integer> offset) {
+    return calculateRangeFilter(toZonedDateTime(logicalStartTime), duration, offset);
+  }
+
+  private SObjectFilterDescriptor(@Nullable ZonedDateTime startTime,
+                                  @Nullable ZonedDateTime endTime) {
+    this.startTime = startTime;
+    this.endTime = endTime;
+  }
+
+  @Nullable
+  public ZonedDateTime getStartTime() {
+    return startTime;
+  }
+
+  @Nullable
+  public ZonedDateTime getEndTime() {
+    return endTime;
+  }
+
+  public boolean isNoOp() {
+    return equals(NO_OP_FILTER_INSTANCE);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(startTime, endTime);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    SObjectFilterDescriptor that = (SObjectFilterDescriptor) o;
+    return Objects.equals(startTime, that.startTime) &&
+      Objects.equals(endTime, that.endTime);
+  }
+
+  @Override
+  public String toString() {
+    return "SObjectFilterDescriptor{" +
+      "startTime=" + startTime +
+      ", endTime=" + endTime +
+      '}';
+  }
+
+  /**
+   * Generates Range or NoOp SObject query filter using provided values.
+   *
+   * @param logicalStartTime application start time
+   * @param duration         filter duration
+   * @param offset           filter offset
+   * @return instance of SObjectFilterDescriptor
+   */
+  private static SObjectFilterDescriptor calculateRangeFilter(ZonedDateTime logicalStartTime,
+                                                              Map<ChronoUnit, Integer> duration,
+                                                              Map<ChronoUnit, Integer> offset) {
+    if (duration.isEmpty() && offset.isEmpty()) {
+      // no filter is required
+      return NO_OP_FILTER_INSTANCE;
+    }
+
+    ZonedDateTime endTime = logicalStartTime;
+    for (Map.Entry<ChronoUnit, Integer> entry : offset.entrySet()) {
+      endTime = endTime.minus(entry.getValue(), entry.getKey());
+    }
+    ZonedDateTime startTime = endTime;
+    for (Map.Entry<ChronoUnit, Integer> entry : duration.entrySet()) {
+      startTime = startTime.minus(entry.getValue(), entry.getKey());
+    }
+    // if duration and offset are zero (Example: 0 Hours)
+    if (startTime.equals(endTime) && endTime.equals(logicalStartTime)) {
+      // no filter is required
+      return NO_OP_FILTER_INSTANCE;
+    }
+    return new SObjectFilterDescriptor(startTime.equals(endTime) ? null : startTime, endTime);
+  }
+
+  /**
+   * Transforms provided time from millis to {@link ZonedDateTime}.
+   *
+   * @param timeInMillis logical start time
+   * @return datetime in UTC timezone
+   */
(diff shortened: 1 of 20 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -42,55 +50,70 @@
...
+  public String getDatetimeBefore() {
+    return datetimeBefore;
   }
 
   protected void validateFilters() {
-    validateSObjectFilter(SalesforceSourceConstants.PROPERTY_DURATION, getDuration());
-    validateSObjectFilter(SalesforceSourceConstants.PROPERTY_OFFSET, getOffset());
+    validateSObjectDatetimeFilter(SalesforceSourceConstants.PROPERTY_DATETIME_AFTER, getDatetimeAfter());
+    validateSObjectDatetimeFilter(SalesforceSourceConstants.PROPERTY_DATETIME_BEFORE, getDatetimeBefore());
+    validateSObjectRangeFilter(SalesforceSourceConstants.PROPERTY_DURATION, getDuration());
```
albertshau: if an exception is thrown (for example, an invalid chrono unit is given, or the integer could not be parsed), this should throw an InvalidConfigPropertyException. Same with the other fields

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -137,11 +161,46 @@ protected String getSObjectQuery(String sObjectName, Schema schema) {
...
+  }
+
+  private Map<ChronoUnit, Integer> extractRangeValue(String rangeValue) {
+    if (Strings.isNullOrEmpty(rangeValue)) {
+      return Collections.emptyMap();
     }
+    return Stream.of(rangeValue.split(","))
+      .map(String::trim)
+      .map(s -> s.split(":", 2))
+      .filter(keyValue -> keyValue.length > 1)
```
albertshau: this ignores invalid input, but I think it's better to throw an exception.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -125,7 +125,7 @@ public void transform(KeyValue<Schema, Map<String, String>> input,
    */
   @Path("getSchema")
```
albertshau: this annotation can be removed. In 6.0.0, the UI now gets the schema based on whatever is set in configurePipeline().

widgets/SalesforceMultiObjects-batchsource.json
```diff
@@ -74,26 +74,52 @@
       "properties": [
         {
           "widget-type": "textbox",
-          "label": "Datetime Filter",
-          "name": "datetimeFilter",
+          "label": "Last Modified After",
+          "name": "datetimeAfter",
           "widget-attributes" : {
-            "placeholder": "Datetime filter in SOQL Date or Date Literal format"
+            "placeholder": "Datetime filter in SOQL Date format"
```
albertshau: I think it's more useful if this shows the user what they can type in. Something like:

YYYY-MM-DDThh:mm:ssZ (ex: 2019-01-01T00:00:00Z)

widgets/SalesforceMultiObjects-batchsource.json
```diff
@@ -74,26 +74,52 @@
...
           }
         },
         {
           "widget-type": "textbox",
+          "label": "Last Modified Before",
+          "name": "datetimeBefore",
+          "widget-attributes" : {
+            "placeholder": "Datetime filter in SOQL Date format"
```
albertshau: same comment here

src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
```diff
@@ -0,0 +1,91 @@
...
+  private final Map<ChronoUnit, Integer> offset;
+
+  public SObjectFilterDescriptor(@Nullable String datetimeAfter,
+                                 @Nullable String datetimeBefore,
+                                 Map<ChronoUnit, Integer> duration,
+                                 Map<ChronoUnit, Integer> offset) {
+    this(null, datetimeAfter, datetimeBefore, duration, offset);
+  }
+
+  public SObjectFilterDescriptor(@Nullable Long startTime,
```
albertshau: I think it's a little cleaner if this is a long instead of nullable Long. Callers should pass in System.currentTimeMillis() instead of null.

src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
```diff
@@ -0,0 +1,91 @@
...
+  public String getDatetimeAfter() {
+    return datetimeAfter;
+  }
+
+  @Nullable
+  public String getDatetimeBefore() {
+    return datetimeBefore;
+  }
+
+  public Map<ChronoUnit, Integer> getDuration() {
```
albertshau: Instead of putting the logic in the util class, I think it's better if this class handles the duration/offset calculations and only exposes getDatetimeBefore() and getDatetimeAfter().

src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public SObjectFilterDescriptor(@Nullable String datetimeAfter,
+                                 @Nullable String datetimeBefore,
+                                 Map<ChronoUnit, Integer> duration,
+                                 Map<ChronoUnit, Integer> offset) {
+    this(null, datetimeAfter, datetimeBefore, duration, offset);
+  }
+
+  public SObjectFilterDescriptor(@Nullable Long startTime,
+                                 @Nullable String datetimeAfter,
```
albertshau: It's better if this class takes ZonedDateTime instead of the string representations, in order to guarantee that anything that uses this class doesn't have to worry about whether these strings are invalid values.

I also think it's better if there is one constructor for before and after, and one for duration offset:

```
public SObjectFilterDescriptor(@Nullable ZonedDateTime datetimeAfter, @Nullable ZonedDateTime datetimeBefore) {
  ...
}

public SObjectFilterDescriptor(long logicalStartTime, Map<ChronoUnit, Integer> duration, Map<ChronoUnit, Integer> offset) {
  ...
}
```

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -42,55 +50,70 @@
...
 
   @Nullable
-  public String getDatetimeFilter() {
-    return datetimeFilter;
+  public String getDatetimeAfter() {
+    return datetimeAfter;
+  }
+
+  @Nullable
+  public String getDatetimeBefore() {
```
albertshau: it's better if these return ZonedDateTime, so that anything that calls it is guaranteed to use valid values.

aielchiieva:
@albertshau addressed code review comments.

src/main/java/io/cdap/plugin/salesforce/SObjectFilterDescriptor.java
albertshau: What I meant is that it's better for that logic to be encapsulated in this class instead of in some utility class. Regardless of whether before/after is given, or duration/offset is given, the final query just cares about a start time and an end time. They are different ways of conveying the same information. 

It is cleaner and more re-usable if callers don't need to handle that logic and only need to handle a nullable start time and nullable end time.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
albertshau: that means the correctness of this class is coupled with the correctness of an unrelated class. If somebody uses it somewhere else in the future, they might not validate the input, resulting in buggy code. It is better to have APIs where correctness is guaranteed if possible.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -189,18 +192,46 @@ private void validateSObjectRangeFilter(String propertyName, Map<ChronoUnit, Int
...
         (o, n) -> n
       ));
   }
+
+  private void validateUnitKeyValue(String propertyName, String rangeValue, String[] keyValue) {
+    if (keyValue.length < 2) {
+      throw new InvalidConfigPropertyException(
+        String.format("Range property '%s' value has invalid format '%s'. "
+                        + "Expected format is <VALUE_1>:<TYPE_1>,<VALUE_2>:<TYPE_2>... . "
+                        + "For example, '1:DAYS,2:HOURS,30:MINUTES'", propertyName, rangeValue), propertyName);
```
albertshau: This format doesn't match what is in the docs (1 days, 2 hours, 30 minutes). I think the space is more friendly than colon and should be used instead

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -189,18 +192,46 @@ private void validateSObjectRangeFilter(String propertyName, Map<ChronoUnit, Int
...
+        keyValue -> parseUnitType(propertyName, keyValue[1]),
+        keyValue -> parseUnitValue(propertyName, keyValue[0]),
         (o, n) -> n
       ));
   }
+
+  private void validateUnitKeyValue(String propertyName, String rangeValue, String[] keyValue) {
+    if (keyValue.length < 2) {
+      throw new InvalidConfigPropertyException(
+        String.format("Range property '%s' value has invalid format '%s'. "
```
albertshau: Range property '%s' value has invalid format -> '%s' has invalid format

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -189,18 +192,46 @@ private void validateSObjectRangeFilter(String propertyName, Map<ChronoUnit, Int
...
+    }
+  }
+
+  private ChronoUnit parseUnitType(String propertyName, String value) {
+
+    try {
+      return ChronoUnit.valueOf(value.trim().toUpperCase());
+    } catch (IllegalArgumentException e) {
+      throw new InvalidConfigPropertyException(
+        String.format("Range property '%s' unit type is invalid '%s'", propertyName, value), e, propertyName);
```
albertshau: '%s' has invalid unit type '%s'

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -189,18 +192,46 @@ private void validateSObjectRangeFilter(String propertyName, Map<ChronoUnit, Int
...
+    }
+  }
+
+  private int parseUnitValue(String propertyName, String value) {
+    try {
+      return Integer.parseInt(value.trim());
+    } catch (NumberFormatException e) {
+      throw new InvalidConfigPropertyException(
+        String.format("Range property '%s' unit value is invalid '%s'", propertyName, value), e, propertyName);
```
albertshau: '%s' has invalid unit value '%d'

aielchiieva:
@albertshau addressed code review comments, please take a look.

PR #33, opened 2019-05-22
CDAP-15313. Salesforce Batch Sink

aonischuk:
Insert records into Salesforce sObjects using Salesforce Bulk API. Dividing the data into separate Salesforce batches and submitting them from separate mappers.

**How Insert to Salesforce Batch API works:**
1. Connect to Bulk API via OAuth2
2. Create a bulk job id
3. Split data into batches
4. Add batches to the job
6. Close job
7. Await every batch completion
8. Check results for every batch and for every record

**For more info:**
https://wiki.cask.co/display/CE/Salesforce+Batch+Sink


nitinmotgi:
Could you please send PR to merge against develop. We should have generally done it against 'develop' but we missed on this one.

aonischuk:
Done

docs/Salesforce-batchsink.md
```diff
@@ -0,0 +1,38 @@
+# Salesforce Batch Sink
+
+
+Description
+-----------
+A batch sink that inserts sObjects into Salesforce.
+Examples of sObjects are opportunities, contacts, accounts, leads, any custom objects, etc.
+
+Plugin can only insert not upsert.
```
albertshau: Currently, only inserts are supported. Upserts are not supported.

docs/Salesforce-batchsink.md
```diff
@@ -0,0 +1,38 @@
...
+
+**SObject Name:** Salesforce object name to insert records into.
+
+**Max records per batch:** Maximum number of records, which plugin puts into a Salesforce batch.
+This value cannot be greater than 10,000. Which is enforced by Salesforce.
+
+**Max bytes per batch:** Maximum number of records, which plugin puts into a Salesforce batch.
+This value cannot be greater than 10,000,000. Which is enforced by Salesforce.
+
+**Error Handling:** Strategy used to handle erroneous records.<br>
```
albertshau, line 36: are the `<br>` needed or can we just have newlines?

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -132,4 +159,77 @@ public static InputStream waitForBatchResults(BulkConnection bulkConnection, Str
...
+   *
+   * @param bulkConnection bulk connection instance
+   * @param job a Salesforce job
+   * @param batchInfoList a list of batches to check
+   * @param ignoreFailures if 'true' not successful row insertion results does not cause an exception.
+   * @throws AsyncApiException if there is an issue checking for batch results
+   * @throws IOException reading csv from Salesforce failed
+   */
+  public static void checkResults(BulkConnection bulkConnection, JobInfo job,
+                            List<BatchInfo> batchInfoList, boolean ignoreFailures)
```
albertshau: nit: indentation

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -132,4 +159,77 @@ public static InputStream waitForBatchResults(BulkConnection bulkConnection, Str
...
     throw new BulkAPIBatchException("Timeout waiting for batch results", info);
   }
+
+  /**
+   * Gets the results of the insert operation for every batch and checks them for errors.
+   *
+   * @param bulkConnection bulk connection instance
+   * @param job a Salesforce job
+   * @param batchInfoList a list of batches to check
+   * @param ignoreFailures if 'true' not successful row insertion results does not cause an exception.
```
albertshau: nit: if true, unsuccessful row insertions do not cause an exception

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -132,4 +159,77 @@ public static InputStream waitForBatchResults(BulkConnection bulkConnection, Str
...
+                            List<BatchInfo> batchInfoList, boolean ignoreFailures)
+    throws AsyncApiException, IOException {
+
+    for (BatchInfo batchInfo : batchInfoList) {
+      CSVReader rdr = new CSVReader(bulkConnection.getBatchResultStream(job.getId(), batchInfo.getId()));
+      List<String> resultHeader = rdr.nextRecord();
+      int resultCols = resultHeader.size();
+
+      List<String> row;
+      while ((row = rdr.nextRecord()) != null) {
```
albertshau, line 187: can you add a comment about what the response is expected to look like? It looks like a csv with a Success column, an Error column, and some other columns that we don't need?

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -132,4 +159,77 @@ public static InputStream waitForBatchResults(BulkConnection bulkConnection, Str
...
+
+  /**
+   * Wait for a job to complete by polling the Bulk API.
+   *
+   * @param bulkConnection BulkConnection used to check results.
+   * @param job The job awaiting completion.
+   * @param batchInfoList List of batches for this job.
+   */
+  public static void awaitCompletion(BulkConnection bulkConnection, JobInfo job,
+                               List<BatchInfo> batchInfoList) {
```
albertshau: nit: indentation

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -132,4 +159,77 @@ public static InputStream waitForBatchResults(BulkConnection bulkConnection, Str
...
+    Set<String> incomplete = batchInfoList
+      .stream()
+      .map(BatchInfo::getId)
+      .collect(Collectors.toSet());
+
+    Awaitility.await()
+      .atMost(GET_BATCH_WAIT_TIME_SECONDS, TimeUnit.SECONDS)
```
albertshau, line 222: where do these constants come from? Is this the max amount of time the Salesforce API takes?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
```diff
@@ -0,0 +1,137 @@
...
+    Schema inputSchema = context.getInputSchema();
+    config.validate(inputSchema);
+
+    context.addOutput(Output.of(config.referenceName, new SalesforceOutputFormatProvider(config)));
+
+    LineageRecorder lineageRecorder = new LineageRecorder(context, config.referenceName);
+    lineageRecorder.createExternalDataset(inputSchema);
+    // Record the field level WriteOperation
+    if (inputSchema.getFields() != null && !inputSchema.getFields().isEmpty()) {
+      lineageRecorder.recordWrite("Write", "Wrote to Salesforce sink",
```
albertshau: would be good to include the object name in the description. Wrote to Salesforce [sObject]

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
```diff
@@ -0,0 +1,137 @@
...
+                                    .collect(Collectors.toList()));
+    }
+  }
+
+  @Override
+  public void transform(StructuredRecord record, Emitter<KeyValue<NullWritable, CSVRecord>> emitter) {
+    List<String> fieldNames = new ArrayList<>();
+    List<Object> values = new ArrayList<>();
+
+    for (Schema.Field field : record.getSchema().getFields()) {
```
albertshau: It's better to have a separate class that is responsible for transforming a StructuredRecord into a CSVRecord. That way it is re-usable in case another sink needs it in the future, and it makes it easy to unit test the conversion behavior.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+    if (fieldSchema.getLogicalType() != null) {
+      Instant instant;
+      switch (logicalType) {
+        case DATE:
+          // convert epoch day to yyyy-mm-dd format
+          return LocalDate.ofEpochDay((Integer) value).toString();
+        case TIMESTAMP_MICROS:
```
albertshau: what about TIMESTAMP_MILLIS and TIME_MILLIS?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
```diff
@@ -0,0 +1,137 @@
...
+          // convert timestamp to ISO 8601 format
+          instant = Instant.ofEpochMilli(TimeUnit.MICROSECONDS.toMillis((Long) value));
+          return instant.toString();
+        case TIME_MICROS:
+          // convert timestamp to HH:mm:ss,SSS
+          instant = Instant.ofEpochMilli(TimeUnit.MICROSECONDS.toMillis((Long) value));
+          return instant.atZone(ZoneId.systemDefault()).toLocalTime();
```
albertshau: I think this is supposed to be UTC, as the time represents the amount of time since midnight. If a different timezone is used, this will result in an incorrect hour.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -0,0 +1,114 @@
...
+        BatchInfo batchInfo = bulkConnection.createBatchFromStream(jobInfo, csvInputStream);
+        batchInfoList.add(batchInfo);
+        LOG.info("Submitted a batch with batchId='{}'", batchInfo.getId());
+      } catch (AsyncApiException e) {
+        throw new RuntimeException("There was issue communicating with Salesforce", e);
+      }
+    }
+
+    if (startNewBatch) {
+      csvBuffer = new CSVBuffer(true);
```
albertshau: instead of creating a new buffer, it would be a little more performant (less garbage collection) if the class had a reset() method.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -0,0 +1,114 @@
...
+      try {
+        BatchInfo batchInfo = bulkConnection.createBatchFromStream(jobInfo, csvInputStream);
+        batchInfoList.add(batchInfo);
+        LOG.info("Submitted a batch with batchId='{}'", batchInfo.getId());
+      } catch (AsyncApiException e) {
+        throw new RuntimeException("There was issue communicating with Salesforce", e);
+      }
+    }
+
+    if (startNewBatch) {
```
albertshau: nit: It's a little cleaner if this method never starts a new batch and doesn't take a boolean, but the write() method starts a new batch and the close() method does not.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -0,0 +1,193 @@
...
+   * According to "Bulk API Limitations" batch cannot contain more than 10,000 records.
+   */
+  private static final long MAX_RECORDS_PER_BATCH_LIMIT = 10_000;
+
+  @Description("Salesforce object name to insert records into.")
+  @Macro
+  private String sObject;
+
+  @Name(PROPERTY_MAX_BYTES_PER_BATCH)
+  @Description("Maximum number of records, which plugin puts into a Salesforce batch.\n" +
```
albertshau: Maximum size in bytes of a batch of records when writing to Salesforce. This value cannot be greater than 10,000,000.

albertshau: I'm pretty sure the numbers are swapped? Would make more since in 10M is the max byte size and 10K is the max records

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -0,0 +1,193 @@
...
+    "This value cannot be greater than 10,000. Which is enforced by Salesforce.")
+  @Macro
+  private String maxBytesPerBatch;
+
+  @Name(PROPERTY_MAX_RECORDS_PER_BATCH)
+  @Description("Maximum number of records, which plugin puts into a Salesforce batch.\n" +
```
albertshau: Maximum number of records to include in a batch when writing to Salesforce. This value cannot be greater than 10,000.

widgets/Salesforce-batchsink.json
```diff
@@ -0,0 +1,108 @@
...
+          "double",
+          "bytes",
+          "string"
+        ],
+        "schema-default-type": "string",
+        "property-watch": "format"
+      }
+    }
+  ]
+}
```
albertshau: all files should end in a newline

docs/Salesforce-batchsink.md
aonischuk: `<br>` creates a line break in markup. While simple line break does not. It's a bit more readable with the newline character. But please let me know what you think.

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
aonischuk: There is a already a comment about this at the top of this class (where constant is defined)
```
  /**
   * Salesforce Bulk API has a limitation, which is 10 minutes per processing of a batch
   */
  private static final long GET_BATCH_WAIT_TIME_SECONDS = 600;
```

This means that Salesforce won't take longer than 10 minutes to process a batch, as per their docs.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
aonischuk: Well the approach is the same as in other Salesforce plugins. We had handling of millis and micros at the same time, but Arina removed millis handling explaining the reasoning at https://github.com/data-integrations/salesforce/pull/14#discussion_r275868661
For now in all plugins we just handle micros not millis.

aonischuk:
Addressed all the comments with either a response or a code change.
@albertshau can you please re-view?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
albertshau: It's true that the UI currently only understands micros (this will likely eventually change) so it makes sense for the source to output micros. However, the sink can take input from any other plugin, and those plugins might use millis. Although we likely won't see this scenario with the current set of plugins, it's still better to handle it in case it gets more prevalent in the future, especially since handling it seems straightforward.

docs/Salesforce-batchsink.md
```diff
@@ -6,7 +6,7 @@ Description
 A batch sink that inserts sObjects into Salesforce.
 Examples of sObjects are opportunities, contacts, accounts, leads, any custom objects, etc.
 
-Plugin can only insert not upsert.
+Currently, only inserts are supported. Upserts are not supported.
```
albertshau, line 9: Let's add a part in the design doc about what would be needed to support updates and upserts, and revisit in a future PR.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -93,16 +92,15 @@ private void submitCurrentBatch(boolean startNewBatch) throws IOException {
...
-      csvBuffer = new CSVBuffer(true);
+      csvBuffer.reset();
     }
   }
 
   @Override
   public void close(TaskAttemptContext taskAttemptContext) throws IOException {
-    submitCurrentBatch(false);
+    submitCurrentBatch();
+    csvBufferSizeCheck.close();
```
albertshau: can these throw IOException?

If so, we need to catch exceptions and make sure we attempt to close both. Also need to make sure that even if these throw an exception we still wait for the job to complete. Probably simpler if these are closed after waiting for the job.

albertshau (commented):
We also discussed a little bit about whether there is some way we can support writing the errors onto HDFS or another filesystem. After more thought, I think it's better not to do a one-off here and to properly add the capability to the sytem (https://issues.cask.co/browse/CDAP-15453).

docs/Salesforce-batchsink.md
aonischuk: Added to design doc.
Also created a jira for the change
https://issues.cask.co/browse/CDAP-15455

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
aonischuk: done

aonischuk:
@albertshau I have done the remaining change. And also merged the commits in one.
Any other changes expected here?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -0,0 +1,119 @@
...
+    try {
+      SalesforceBulkUtil.awaitCompletion(bulkConnection, jobInfo, batchInfoList);
+      SalesforceBulkUtil.checkResults(bulkConnection, jobInfo, batchInfoList, errorHandling.equals(ErrorHandling.SKIP));
+    } catch (AsyncApiException e) {
+      throw new RuntimeException("There was issue communicating with Salesforce", e);
+    }
+
+    try {
+      csvBufferSizeCheck.close();
+    } catch(IOException ex) {
```
albertshau: missing space after catch

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -0,0 +1,119 @@
...
+    submitCurrentBatch();
+
+    try {
+      SalesforceBulkUtil.awaitCompletion(bulkConnection, jobInfo, batchInfoList);
+      SalesforceBulkUtil.checkResults(bulkConnection, jobInfo, batchInfoList, errorHandling.equals(ErrorHandling.SKIP));
+    } catch (AsyncApiException e) {
+      throw new RuntimeException("There was issue communicating with Salesforce", e);
+    }
+
+    try {
```
albertshau: this should happen even if an AsyncApiException is caught above. Probably should have it in a finally block from above.

albertshau:
Still have a pending comment about handling timestamp millis and time millis

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
aonischuk: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
aonischuk: fixed

aonischuk:
@albertshau done.
fixed new comments. And added handling of millis.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceBatchSink.java
aonischuk: done

aonischuk:
Have resolved all the points.
@albertshau is it possible to merge at this point?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+          // convert timestamp to ISO 8601 format
+          instant = Instant.ofEpochMilli(TimeUnit.MICROSECONDS.toMillis((Long) value));
+          return instant.toString();
+        case TIME_MICROS:
+          // convert timestamp to HH:mm:ss,SSS
+          instant = Instant.ofEpochMilli(TimeUnit.MICROSECONDS.toMillis((Long) value));
+          return instant.atZone(ZoneOffset.UTC).toLocalTime();
+        case TIMESTAMP_MILLIS:
+          // convert timestamp to ISO 8601 format
+          instant = Instant.ofEpochMilli(TimeUnit.MILLISECONDS.toMillis((Long) value));
```
albertshau: TimeUnit.MILLISECONDS.toMillis() is a no-op and not needed.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+          return instant.toString();
+        case TIME_MILLIS:
+          // convert timestamp to HH:mm:ss,SSS
+          instant = Instant.ofEpochMilli(TimeUnit.MILLISECONDS.toMillis((Long) value));
```
albertshau: same here

albertshau (approved):
couple minor comments, otherwise lgtm. Fix and I'll merge

aonischuk:
@albertshau I have the fixed the given comments.
Can we please merge?

albertshau (approved):
lgtm

PR #34, opened 2019-05-22
CDAP-15236: Bulk API limitations handling

aielchiieva:
1. Support aggregate function calls
2. Support sub-queries
3. Map to record transformer refactoring
4. Support CDAP array and record types
5. Transform sub-queries to CDAP array of records

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
@@ -26,18 +26,20 @@ import io.cdap.plugin.salesforce.parser.SalesforceQueryParser;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
- * Contains information about SObject, including its name and list of fields.
+ * Contains information about SObject, including its name, list of fields and sub-query elements.
  * Can be obtained from SOQL query or from SObject name.
  */
 public class SObjectDescriptor {
 
   private final String name;
   private final List<FieldDescriptor> fields;
+  private final List<SObjectDescriptor> childSObjects;
 
   /**
    * Connects to Salesforce, gets describe result for the given sObject name and stores
@@ -53,7 +55,7 @@ public class SObjectDescriptor {
                                            AuthenticatorCredentials credentials, Set<FieldType> typesToSkip)
     throws ConnectionException {
     PartnerConnection partnerConnection = SalesforceConnectionUtil.getPartnerConnection(credentials);
-    SObjectsDescribeResult describeResult = new SObjectsDescribeResult(
+    SObjectsDescribeResult describeResult = SObjectsDescribeResult.of(
       partnerConnection, Collections.singletonList(name));
     List<FieldDescriptor> fields = describeResult.getFields().stream()
       .filter(field -> !typesToSkip.contains(field.getType()))
@@ -80,8 +82,13 @@ public class SObjectDescriptor {
   }
 
   public SObjectDescriptor(String name, List<FieldDescriptor> fields) {
+    this(name, fields, Collections.emptyList());
+  }
+
+  public SObjectDescriptor(String name, List<FieldDescriptor> fields, List<SObjectDescriptor> childSObjects) {
     this.name = name;
     this.fields = new ArrayList<>(fields);
+    this.childSObjects = childSObjects;
   }
 
   public String getName() {
@@ -89,30 +96,43 @@ public class SObjectDescriptor {
   }
 
   /**
-   * Collects sObject names needed to be described in order to obtains field type information.
+   * Collects unique combinations of featured SObjects.
    *
-   * @return list of sObject names
+   * For the query
+   * `SELECT DandbCompany.Name, (SELECT FirstName, LastName, Owner.UserRole.Id FROM Contacts) FROM Account`
+   * there are three unique combinations of features SObjects:
+   * Name -> FeaturedSObjects{relationship=null, references=[DandbCompany]}
+   * FirstName, LastName -> FeaturedSObjects{relationship='Contacts', references=[]}
+   * Id -> FeaturedSObjects{relationship='Contacts', references=[Owner, UserRole]}
+   *
+   * @return unique combinations of featured SObjects
    */
-  public Set<String> getAllParentObjects() {
-    Set<String> parents = fields.stream()
+  public Set<FeaturedSObjects> getFeaturedSObjects() {
+    Set<FeaturedSObjects> result = getFeaturedSObjects(null, getFields());
+
+    for (SObjectDescriptor child : getChildSObjects()) {
+      result.add(new FeaturedSObjects(child.getName(), Collections.emptyList()));
+      result.addAll(getFeaturedSObjects(child.getName(), child.getFields()));
+    }
+
+    return result;
+  }
+
+  private Set<FeaturedSObjects> getFeaturedSObjects(String name, List<FieldDescriptor> fields) {
+    return fields.stream()
       .filter(FieldDescriptor::hasParents)
-      .map(FieldDescriptor::getLastParent)
+      .map(f -> new FeaturedSObjects(name, f.getParents()))
       .collect(Collectors.toSet());
-
-    // add top level sObject for fields that don't have parents
-    parents.add(name);
-
-    return parents;
   }
 
   /**
-   * Collects all field names, for fields with parents includes parents separated by dot.
+   * Collects all field names, for fields with parents, parents are separated by dot.
    *
    * @return list of field names
    */
   public List<String> getFieldsNames() {
     return fields.stream()
-      .map(FieldDescriptor::getFullName)
+      .map(FieldDescriptor::getQueryName)
       .collect(Collectors.toList());
   }
 
@@ -120,30 +140,86 @@ public class SObjectDescriptor {
     return fields;
   }
 
+  public List<SObjectDescriptor> getChildSObjects() {
+    return childSObjects;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, fields, childSObjects);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    SObjectDescriptor that = (SObjectDescriptor) o;
+    return name.equals(that.name)
+      && fields.equals(that.fields)
+      && childSObjects.equals(that.childSObjects);
+  }
+
   @Override
   public String toString() {
-    return "SObjectDescriptor{" + "name='" + name + '\'' + ", fields=" + fields + '}';
+    return "SObjectDescriptor{"
+      + "name=" + name
+      + ", fields=" + fields
+      + ", childSObjects=" + childSObjects
+      + '}';
   }
 
   /**
    * Contains information about field, including list of parents if present.
+   * Can contain field alias and function type if field is used with function call.
    */
   public static class FieldDescriptor {
 
     private final Field field;
     private final List<String> parents;
+    @Nullable
+    private final String alias;
+    private final SalesforceFunctionType functionType;
 
     public FieldDescriptor(Field field) {
       this.field = field;
       this.parents = new ArrayList<>();
+      this.alias = null;
+      this.functionType = SalesforceFunctionType.NONE;
     }
 
-    public FieldDescriptor(List<String> nameParts) {
+    /**
+     * Creates field descriptor instance based on given field information.
+     * <ul>
+     * <li>`nameParts` contains list of field names.
+     * For simple field it will contain singleton list with field name only (SELECT Id FROM Opportunity).
+     * For reference fields each list element will correspond to the element in the field name
+     * initially separated by dot (SELECT Account.Name FROM Contact).</li>
+     * <li>`alias` represents field alternative name given in the query.
+     * Only applicable for queries with aggregate function calls.
+     * If field does not have alias, its value is null.</li>
+     * <li>`functionType` indicates function type for the fields used in function calls.
+     * If field was not used in function call,
+     * its function type will be {@link SalesforceFunctionType#NONE}.</li>
+     * </ul>
+     *
+     * @param nameParts all field name parts
+     * @param alias field alias
+     * @param functionType function type for the fields used in function calls
+     */
+    public FieldDescriptor(List<String> nameParts, String alias,
+                           SalesforceFunctionType functionType) {
       Preconditions.checkState(nameParts != null && !nameParts.isEmpty(),
         "Given list of name parts must contain at least one element");
       this.parents = new ArrayList<>(nameParts);
       this.field = new Field();
       field.setName(parents.remove(nameParts.size() - 1));
+
+      this.alias = alias;
+      this.functionType = functionType;
     }
 
     public String getName() {
@@ -165,7 +241,12 @@ public class SObjectDescriptor {
     }
 
     /**
-     * Checks if field has parents.
+     * Checks if field has parents. Only reference fields have parents.
+     * Example:
+     * <ul>
+     * <li>`SELECT Id from Opportunity`: `Id` field does not have parents.</li>
+     * <li>`SELECT Account.Name FROM Contact`: `Name` field has one parent `Account`.</li>
+     * </ul>
      *
      * @return true if field has at least one parent, false otherwise
      */
@@ -177,14 +258,10 @@ public class SObjectDescriptor {
       return parents;
     }
 
-    /**
-     * Return last parent of the field.
-     * Primary used to obtain describe result from Salesforce.
-     *
-     * @return last parent if field has parents, null otherwise
-     */
-    public String getLastParent() {
-      return hasParents() ? parents.get(parents.size() - 1) : null;
+    public String getParentsPath(List<String> topLevelParents) {
+      List<String> allParents = new ArrayList<>(topLevelParents);
+      allParents.addAll(parents);
+      return String.join(SalesforceConstants.REFERENCE_NAME_DELIMITER, allParents);
     }
 
     @Nullable
@@ -192,10 +269,155 @@ public class SObjectDescriptor {
       return field.getType();
     }
 
+    public List<String> getNameParts() {
+      List<String> nameParts = new ArrayList<>(parents);
+      nameParts.add(field.getName());
+      return nameParts;
+    }
+
+    /**
+     * Checks if field has alias.
+     *
+     * @return true if field has alia, false otherwise
+     */
+    public boolean hasAlias() {
+      return alias != null;
+    }
+
+    /**
+     * Returns field alias if it has one.
+     *
+     * @return field alias
+     */
+    public String getAlias() {
+      return alias;
+    }
+
+    /**
+     * If fields name was used with alias, returns field alias,
+     * otherwise returns full field name with parents if any.
+     *
+     * @return field query name
+     */
+    public String getQueryName() {
+      return hasAlias() ? alias : getFullName();
+    }
+
+    /**
+     * If field was used with function call returns used function type.
+     *
+     * @return function type
+     */
+    public SalesforceFunctionType getFunctionType() {
+      return functionType;
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(field.getName(), field.getType(), field.isNillable(),
+        parents, alias, functionType);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      FieldDescriptor that = (FieldDescriptor) o;
+      return Objects.equals(field.getName(), that.field.getName())
+        && field.getType() == that.field.getType()
+        && field.isNillable() == that.field.isNillable()
+        && parents.equals(that.parents)
+        && Objects.equals(alias, that.alias)
+        && functionType == that.functionType;
+    }
+
     @Override
     public String toString() {
-      return "FieldDescriptor{" + "name='" + field.getName() + '\'' + ", parents=" + parents + '}';
+      return "FieldDescriptor{"
+        + "fieldName=" + field.getName()
+        + ", fieldType=" + field.getType()
+        + ", nillable=" + field.isNillable()
+        + ", parents=" + parents
+        + ", alias=" + alias
+        + ", functionType=" + functionType
+        + '}';
     }
   }
 
+  /**
+   * Describes combination of SObjects used in the query for the field.
+   * For the query
+   * `SELECT DandbCompany.Name, (SELECT FirstName, LastName, Owner.UserRole.Id FROM Contacts) FROM Account`
+   * there are three unique combinations of features SObjects:
+   * Name -> FeaturedSObjects{relationship=null, references=[DandbCompany]}
+   * FirstName, LastName -> FeaturedSObjects{relationship='Contacts', references=[]}
+   * Id -> FeaturedSObjects{relationship='Contacts', references=[Owner, UserRole]}
+   */
+  public static class FeaturedSObjects {
+    @Nullable
+    private final String relationship;
+    private final List<String> references;
+
+    public FeaturedSObjects(String relationship, List<String> references) {
+      this.relationship = relationship;
+      this.references = references == null ? Collections.emptyList() : references;
+    }
+
+    public String getRelationship() {
+      return relationship;
+    }
+
+    public List<String> getReferences() {
+      return references;
+    }
+
+    public boolean hasRelationship() {
+      return relationship != null;
+    }
+
+    public boolean hasReferences() {
+      return !references.isEmpty();
+    }
+
+    public String getFullName(String topLevelName) {
+      List<String> names = new ArrayList<>();
+      names.add(topLevelName);
+      if (hasRelationship()) {
+        names.add(relationship);
+      }
+      if (hasReferences()) {
+        names.addAll(references);
+      }
+      return String.join(SalesforceConstants.REFERENCE_NAME_DELIMITER, names);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(relationship, references);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      FeaturedSObjects obj = (FeaturedSObjects) o;
+      return Objects.equals(relationship, obj.relationship) && Objects.equals(references, obj.references);
+    }
+
+    @Override
+    public String toString() {
+      return "FeaturedSObjects{"
+        + "relationship=" + relationship
+        + ", references=" + references
+        + '}';
+    }
+  }
 }
--- a/src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
@@ -15,8 +15,8 @@
  */
 package io.cdap.plugin.salesforce;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
+import com.sforce.soap.partner.ChildRelationship;
 import com.sforce.soap.partner.DescribeSObjectResult;
 import com.sforce.soap.partner.Field;
 import com.sforce.soap.partner.PartnerConnection;
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Retrieves {@link DescribeSObjectResult}s for the given sObjects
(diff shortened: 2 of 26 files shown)


nitinmotgi:
Could you change the branch to develop. All repos generally use 'develop'. We had missed this one. Am sorry.

aielchiieva:
@nitinmotgi no problem, done.

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
```diff
@@ -59,13 +59,24 @@ public static String createSObjectQuery(List<String> fields, String sObjectName,
...
   }
 
+  /**
+   * Checks if given query fall under restricted query category,
+   * which means it cannot be processed using Bulk API.
+   *
+   * @param query SOQL query
+   * @return true if query is restricted, false otherwise
+   */
+  public static boolean isRestrictedQuery(String query) {
```
albertshau: why wouldn't the caller just call SalesforceQueryParser.isRestrictedQuery?

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
```diff
@@ -151,13 +151,21 @@ public static Schema getSchemaWithFields(SObjectDescriptor sObjectDescriptor, SO
...
-      String parent = fieldDescriptor.hasParents() ? fieldDescriptor.getLastParent() : sObjectDescriptor.getName();
-      Field field = describeResult.getField(parent, fieldDescriptor.getName());
-      if (field == null) {
-        throw new IllegalArgumentException(
-          String.format("Field '%s' is absent in Salesforce describe result", fieldDescriptor.getFullName()));
+
+      SalesforceFunctionType functionType = fieldDescriptor.getFunctionType();
+      Schema fieldSchema = null;
+      if (!functionType.isConstant()) {
+        String parent = fieldDescriptor.hasParents() ? fieldDescriptor.getLastParent() : sObjectDescriptor.getName();
```
albertshau: I'm not clear on what a field's parents are. Can you add some comments with an example?

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -57,6 +59,30 @@ public SObjectDescriptor visitStatement(SOQLParser.StatementContext ctx) {
...
+      }
+    } else {
+      List<String> unAliasedFields = fields.stream()
+        .filter(fieldDescriptor -> !fieldDescriptor.hasAlias())
+        .filter(SObjectDescriptor.FieldDescriptor::hasParents)
+        .map(SObjectDescriptor.FieldDescriptor::getFullName)
+        .collect(Collectors.toList());
+
+      if (!unAliasedFields.isEmpty()) {
+        throw new SOQLParsingException("Reference fields must have alias when used with GROUP BY clause. "
```
albertshau: Fields must have an alias when using a GROUP BY clause.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -57,6 +59,30 @@ public SObjectDescriptor visitStatement(SOQLParser.StatementContext ctx) {
...
 
+    if (ctx.fromStatement().GROUP() == null) {
+      List<String> aliasedFields = fields.stream()
+        .filter(fieldDescriptor -> fieldDescriptor.getFunctionType() == SalesforceFunctionType.UNDEFINED)
+        .filter(SObjectDescriptor.FieldDescriptor::hasAlias)
+        .map(SObjectDescriptor.FieldDescriptor::getFullName)
+        .collect(Collectors.toList());
+
+      if (!aliasedFields.isEmpty()) {
+        throw new SOQLParsingException("Field aliasing is not allowed without GROUP BY clause. "
```
albertshau: Field aliasing is not allowed except when using a GROUP BY clause.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -57,6 +59,30 @@ public SObjectDescriptor visitStatement(SOQLParser.StatementContext ctx) {
...
     }
 
+    if (ctx.fromStatement().GROUP() == null) {
+      List<String> aliasedFields = fields.stream()
+        .filter(fieldDescriptor -> fieldDescriptor.getFunctionType() == SalesforceFunctionType.UNDEFINED)
+        .filter(SObjectDescriptor.FieldDescriptor::hasAlias)
+        .map(SObjectDescriptor.FieldDescriptor::getFullName)
+        .collect(Collectors.toList());
+
+      if (!aliasedFields.isEmpty()) {
```
albertshau, line 65: how come we don't allow aliasing?

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -192,9 +219,77 @@ public FieldType getFieldType() {
...
+    public String getQueryName() {
+      return hasAlias() ? alias : getFullName();
+    }
+
+    /**
+     * If field was used with function call returns used function type.
+     *
+     * @return function type
+     */
+    public SalesforceFunctionType getFunctionType() {
```
albertshau, line 311: what is the function type for non-function fields? Can this be null?

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -106,44 +107,70 @@ public String getName() {
...
 
     public FieldDescriptor(Field field) {
       this.field = field;
       this.parents = new ArrayList<>();
+      this.alias = null;
+      this.functionType = SalesforceFunctionType.UNDEFINED;
     }
 
-    public FieldDescriptor(List<String> nameParts) {
+    public FieldDescriptor(List<String> nameParts, String alias, SalesforceFunctionType functionType) {
```
albertshau: I don't understand what name parts are. Can you add some documentation with examples.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
albertshau: I see later on that it's undefined. Doesn't this mean you can't tell the difference between a field without a function and a field that uses a function that isn't in the FunctionType map?

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -81,28 +87,53 @@ public static SObjectDescriptor fromQuery(String query) {
...
+    this.childSObjects = childSObjects;
   }
 
   public String getName() {
     return name;
   }
 
   /**
-   * Collects sObject names needed to be described in order to obtains field type information.
+   * Collects sObject names needed to be described in order to obtain field type information.
```
albertshau: I'm not sure I understand what this method is doing. What are the parents of an SObjectDescriptor when the descriptor only has children? Can you include some example outputs based on different child objects?

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -157,19 +160,26 @@ public SObjectDescriptor visitStatement(SOQLParser.StatementContext ctx) {
...
+          + ctx.getText());
+      }
+      return null;
     }
 
     @Override
     public SObjectDescriptor.FieldDescriptor visitFunctionCall(SOQLParser.FunctionCallContext ctx) {
-      SOQLParser.FunctionContext function = ctx.function();
+      if (subQueryScope) {
+        throw new SOQLParsingException("Function call is not supported in sub-queries: " + ctx.getText());
```
albertshau: Function calls are not supported in sub-queries:

albertshau (commented):
have some comments on the aggregates piece.

Started reviewing the nested queries piece and have some questions about the overall approach. From the design doc, it looks like it's handling queries of the form:

SELECT Id,Account.Name, (SELECT Id,Name,Contact.LastName FROM Account.Contacts) FROM Account

Are other types of subqueries not supported in SOQL? When I think of a subquery in SQL, I usually think of it being used in the FROM clause, like:

Select A.id A.name from (select ...) as A

Also, it looks like the entire subquery is being represented as an array of maps in CDAP? It seems more natural for it to be a flat record, with fields from the subquery moved to the top level. 

There also seems to be a lot of custom query parsing and examination happening here. Salesforce doesn't have an API that tells you the schema that a given query will return? I worry about correctness of all of this, as it means much of the query logic needs to be replicated in this plugin.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: Added new function type `NONE` to separate these two concepts.

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
aielchiieva: Replaced with direct method call.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
aielchiieva: Because Salesforce allows aliases only for aggregated calls. Example of the error returned from Salesforce:
`only aggregate expressions use field aliasing`.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
aielchiieva: Changed.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
aielchiieva: Changed.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
aielchiieva: Changed.

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
aielchiieva: Added example in the javadoc for the `hasParents` method:
```
    /**
     * Checks if field has parents. Only reference fields have parents.
     * Example:
     * <ul>
     * <li>`SELECT Id from Opportunity`: `Id` field does not have parents.</li>
     * <li>`SELECT Account.Name FROM Contact`: `Name` field has one parent `Account`.</li>
     * </ul>
     *
     * @return true if field has at least one parent, false otherwise
     */
```

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: Added java doc for the constructor:
```
    /**
     * Creates field descriptor instance based on given field information.
     * <ul>
     * <li>`nameParts` contains list of field names.
     * For simple field it will contain singleton list with field name only (SELECT Id FROM Opportunity).
     * For reference fields each list element will correspond to the element in the field name
     * initially separated by dot (SELECT Account.Name FROM Contact).</li>
     * <li>`alias` represents field alternative name given in the query.
     * Only applicable for queries with aggregate function calls. 
     * If field does not have alias, its value is null.</li>
     * <li>`functionType` indicates function type for the fields used in function calls.
     * If field was not used in function call,
     * its function type will be {@link SalesforceFunctionType#NONE}.</li>
     * </ul>
     *
     * @param nameParts all field name parts
     * @param alias field alias
     * @param functionType function type for the fields used in function calls
     */
```

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: Added examples in the method javadoc:
```
  /**
   * Collects sObject names needed to be described in order to obtain field type information.
   * Examples:
   * <ul>
   *  <li>
   *    <ol>
   *      <li>Query: SELECT Id from Opportunity</li>
   *      <li>Output: [{Opportunity=TOP_LEVEL}]</li>
   *      <li>Description: `Id` field is simple and does not have parents,
   *      thus to describe this field (define its type), 
   *      only SObject indicated in the FROM clause is needed,
   *      i.e. top level SObject: {Opportunity=TOP_LEVEL}.</li>
   *    </ol>
   *  </li>
   *  <li>
   *    <ol>
   *      <li>Query: SELECT Id, Account.Name FROM Contact</li>
   *      <li>Output: [{Contact=TOP_LEVEL}, {Contact=TOP_LEVEL, Account=REFERENCE}]</li>
   *      <li>Description: `Id` field is simple and does not have parents,
   *      only top level SObject is needed to define its type ({Contact=TOP_LEVEL}).
   *      `Name` is reference field, to define its type top level and reference SObjects
   *      are needed: {Contact=TOP_LEVEL, Account=REFERENCE}.</li>
   *    </ol>
   *  </li>
   *  <li>
   *    <ol>
   *      <li>Query: SELECT Name, (SELECT LastName FROM Contacts) FROM Account</li>
   *      <li>Output: [{Account=TOP_LEVEL}, {Account=TOP_LEVEL, Contacts=RELATIONSHIP}]</li>
   *      <li>Description: `Name` field is simple and does not have parents,
   *      only top level SObject is needed to define its type ({Account=TOP_LEVEL}).
   *      `LastName` is field indicated in sub-query, to define its type top level
   *      and relationship SObjects are needed: {Account=TOP_LEVEL, Contacts=RELATIONSHIP}.</li>
   *    </ol>
   *  </li>
   * </ul>
   *
   * @return set of unique sObject names combinations and their types
   */
```

aielchiieva (commented):
> Are other types of subqueries not supported in SOQL? When I think of a subquery in SQL, I usually think of it being used in the FROM clause, like: Select A.id A.name from (select ...) as A

No, Salesforce does not other types of sub-queries.

https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_relationships_query_using.htm

> Also, it looks like the entire subquery is being represented as an array of maps in CDAP? It seems more natural for it to be a flat record, with fields from the subquery moved to the top level.

Sub-query has one to many relationship with top level fields, so flattening will duplicate top level fields values. Salesforce outputs results for such queries in array of records form as well.

Example of output for `SELECT Id, Account.Name, (SELECT Id, Name, Contact.LastName FROM Account.Contacts) FROM Account `:

```
{
   "totalSize":1,
   "records":[
      {
         "Id":"0011i000005o4KzAAI",
         "Name":"GenePoint",
         "Contacts":{
            "records":[
               {
                  "Id":"0031i000005c49GAAQ",
                  "Name":"Edna Frank",
                  "LastName":"Frank"
               },
               {
                  "Id":"0031i000005c49GAAZ",
                  "Name":"John Brown",
                  "LastName":"Brown"
               }
            ]
         }
      }
   ]
}
```

> There also seems to be a lot of custom query parsing and examination happening here. Salesforce doesn't have an API that tells you the schema that a given query will return? I worry about correctness of all of this, as it means much of the query logic needs to be replicated in this plugin.

Unfortunately Salesforce does not have API to define SOQL schema. The only way to define SOQL schema is to parse given SOQL, define used SObjects and their fields and then describe them.

aielchiieva:
@albertshau addressed code review comments, please take a look.

albertshau:
I see, I had a wrong understanding of what sub-queries are. Thanks for the explanation.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
albertshau: Thanks this javadoc helps a lot

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -80,70 +87,184 @@ public static SObjectDescriptor fromQuery(String query) {
...
     }
 
-    public FieldDescriptor(List<String> nameParts) {
+    /**
+     * Creates field descriptor instance based on given field information.
+     * <ul>
+     * <li>`nameParts` contains list of field names.
+     * For simple field it will contain singleton list with field name only (SELECT Id FROM Opportunity).
+     * For reference fields each list element will correspond to the element in the field name
+     * initially separated by dot (SELECT Account.Name FROM Contact).</li>
```
albertshau, line 200: so in this case nameParts is ["Account", "Name"]? Does this mean nameParts is also 1 or 2 elements long or can it be more? If it can be 3+, what is an example, and what does the order in the list mean?

Basically I'm wondering whether we can pass in one String for name and another nullable String for the parent.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -80,70 +87,184 @@ public static SObjectDescriptor fromQuery(String query) {
...
+   *      <li>Description: `Id` field is simple and does not have parents,
+   *      only top level SObject is needed to define its type ({Contact=TOP_LEVEL}).
+   *      `Name` is reference field, to define its type top level and reference SObjects
+   *      are needed: {Contact=TOP_LEVEL, Account=REFERENCE}.</li>
+   *    </ol>
+   *  </li>
+   *  <li>
+   *    <ol>
+   *      <li>Query: SELECT Name, (SELECT LastName FROM Contacts) FROM Account</li>
+   *      <li>Output: [{Account=TOP_LEVEL}, {Account=TOP_LEVEL, Contacts=RELATIONSHIP}]</li>
```
albertshau: Is the top level object always present in any map returned? And is each map at most 2 items? I'm wondering if this can be more clearly represented as a new class that looks something like:

```
public class QuerySObjects {
  private final String topLevelObject;
  private final Set<String> relationships;
  private final Set<String> references;
}
```

When there are nested collections like a `Set<Map<String, SObjectType>>`, it's hard to understand what each object in each collection represents.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
```diff
@@ -108,6 +134,92 @@ private void addSObjectDescribe(DescribeSObjectResult sObjectDescribe) {
...
+   */
+  private void addSObjectDescribe(PartnerConnection connection,
+                                  Map<String, SObjectDescriptor.SObjectType> names,
+                                  Map<String, DescribeSObjectResult> cache) throws ConnectionException {
+    DescribeSObjectResult previousDescribe = null;
+    for (Map.Entry<String, SObjectDescriptor.SObjectType> entry : names.entrySet()) {
+      String name = entry.getKey();
+      switch (entry.getValue()) {
+        case TOP_LEVEL:
+          previousDescribe = describe(connection, name, cache);
```
albertshau: It looks like this method requires that the map is an ordered map, which I believe is guaranteed with the creation of a LinkedHashMap somewhere else in the code, but could easily be broken at some point in the future.

I think this method would be less brittle and easier to understand if the top level object is passed in as one argument and all the collection of references and relationships are passed in as separate arguments. This would remove the need for this previousDescribe variable, remove the need for a switch on type, and remove the assumption about the map that is given.

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
```diff
@@ -147,20 +148,49 @@ public static void checkCompatibility(Schema actualSchema, Schema providedSchema
...
+      Schema fieldSchema = null;
+      if (!functionType.isConstant()) {
+        List<String> allTopLevelParents = new ArrayList<>(topLevelParents);
+        allTopLevelParents.add(sObjectDescriptor.getName());
+        String parentsPath = fieldDescriptor.getParentsPath(allTopLevelParents);
+        Field field = describeResult.getField(parentsPath, fieldDescriptor.getName());
+
+        if (field == null) {
+          throw new IllegalArgumentException(
+            String.format("Field '%s' is absent in Salesforce describe result", fieldDescriptor.getFullName()));
```
albertshau: include the name of the sobject in the error message, something like:

Field '%s' does not exist in Salesforce object '%s'.

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
```diff
@@ -147,20 +148,49 @@ public static void checkCompatibility(Schema actualSchema, Schema providedSchema
...
+        fieldSchema = getCdapFieldSchema(field);
       }
-      Schema.Field schemaField = Schema.Field.of(fieldDescriptor.getFullName(), getCdapFieldSchema(field));
+
+      Schema queryFieldSchema = functionType.getSchema(fieldSchema);
+      Schema.Field schemaField = Schema.Field.of(fieldDescriptor.getQueryName(), queryFieldSchema);
       schemaFields.add(schemaField);
     }
 
+    for (SObjectDescriptor childSObject : sObjectDescriptor.getChildSObjects()) {
```
albertshau, line 211: Can you add some comments here about what type of query generates child objects? This will make it more clear why that field has a type that is an array of records.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
albertshau: I see. Can you add a comment? 

In general, it would be good to add comments that relate to how the Salesforce API is defined or how it behaves. You now know a lot of Salesforce specific information that the next person who touches the code might not know.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/MapToRecordTransformer.java
```diff
@@ -21,72 +21,90 @@
...
       case DOUBLE:
-        return Double.parseDouble(value);
+        return Double.parseDouble(castValue(value, fieldName, String.class));
       case STRING:
         return value;
+      case RECORD:
+        Map<String, String> recordValues = castGeneric(castValue(value, fieldName, Map.class));
+        StructuredRecord.Builder nestedBuilder = StructuredRecord.builder(fieldSchema);
+        Objects.requireNonNull(fieldSchema.getFields()).forEach(
+          field -> transformRecord(fieldSchema, recordValues, nestedBuilder));
```
albertshau, line 83: there can only be one level of nesting right? No subquery within a subquery?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/MapToRecordTransformer.java
```diff
@@ -21,72 +21,90 @@
...
+  private <T> T castValue(Object value, String fieldName, Class<T> clazz) {
+    if (clazz.isAssignableFrom(value.getClass())) {
+      return clazz.cast(value);
+    }
+    throw new UnexpectedFormatException(
+      String.format("Field '%s' is not of expected type '%s'", fieldName, clazz.getSimpleName()));
   }
 
+  @SuppressWarnings("unchecked")
+  private <T> T castGeneric(Object value) {
```
albertshau, line 107: this doesn't seem any different than just casting. Or is the purpose to avoid warnings?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -135,4 +132,15 @@ private BulkConnection getBulkConnection(Configuration conf) {
     }
   }
 
+  private RecordReader<Schema, Map<String, ?>> getDelegateRecordReader(String query, Schema schema) {
+    if (SalesforceQueryParser.isRestrictedQuery(query)) {
+      LOG.warn("Restricted SOQL query detected switching to SOAP reader. Performance can be significantly impacted");
```
albertshau: This seems like it would be a confusing message for a user to see. If I saw this I would wonder what a restricted query is and how significant will the performance be impacted.

I think it's better if the message said why it is restricted and describes what exactly that means. For example, something like:

The SOQL query uses an aggregate function. Reads will be performed serially and not in parallel.

This is also expected behavior, so I would make it an info message and not a warning.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -135,4 +132,15 @@ private BulkConnection getBulkConnection(Configuration conf) {
...
+      return new SalesforceSoapRecordReader(schema, query, new SoapRecordToMapTransformer());
+    }
+    if (SalesforceQueryUtil.isQueryUnderLengthLimit(query)) {
+      return new SalesforceBulkRecordReader(schema);
+    }
+    LOG.warn("Wide SOQL query detected partially switching to SOAP reader. Performance may be impacted");
```
albertshau: The SOQL query is a wide query. An additional SOAP request will be performed for each record.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSoapRecordReader.java
```diff
@@ -0,0 +1,143 @@
...
+    LOG.debug("Executing Salesforce SOAP query: '{}'", query);
+
+    Configuration conf = taskAttemptContext.getConfiguration();
+    try {
+      AuthenticatorCredentials credentials = SalesforceConnectionUtil.getAuthenticatorCredentials(conf);
+      partnerConnection = SalesforceConnectionUtil.getPartnerConnection(credentials);
+      sObjectDescriptor = SObjectDescriptor.fromQuery(query);
+      queryResult = partnerConnection.query(query);
+    } catch (ConnectionException e) {
+      throw new RuntimeException("Cannot create Salesforce SOAP connection", e);
```
albertshau, line 78: nit: I think the initialize method can throw IOException? That seems like a slightly better exception to use.

albertshau (commented):
some questions/comments about replacing java collections with custom classes.

In the future, it would be good to split this type of change into separate PRs. One for aggregates, another for subqueries.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: Yes, it can be 3+. For example, `SELECT Name, (SELECT LastName, Owner.UserRole.Id FROM Contacts) FROM Account`.
`Id` column will have 3 parents: `Contacts`, `Owner`, `UserRole`.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
aielchiieva: Passing separately references and relationships wont work since we can have more that two levels of nesting. 
As well as the need of previous describe result variable. Describing references and relationship in Salesforce is quite complex process. Since Salesforce is having aliases for references and relationships and depending on the type different methods to find out real SObject name.

Example:
`SELECT Name, (SELECT Owner.Id FROM Contacts) FROM Account`
To get `Id` column type, we need to describe Account, Contacts and Owner.
`Account` is top level SObject so we can describe by name provided in the `FROM` clause.
`Contacts` is relationship name which refers to `Contact` SObject, we need to use `Account` describe result and look through its child-relationship names to find out that `Contacts` refers to `Contact`. After that we can describe `Contact` SObject. 
Next we need to describe `Owner`. `Owner` is reference name, so we need to use `Contact` describe result and look through its reference names to find out that `Owner` refers to `User` SObject. Finally, we describe `User` SObject and can get information about `Id` column type.

Anyway, introduced `FeaturedSObject` class, now we will pass collection of this class instead of set of maps.

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
aielchiieva: Done.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
aielchiieva: Done.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/MapToRecordTransformer.java
aielchiieva: Correct.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/MapToRecordTransformer.java
aielchiieva: Yes, just to avoid warnings in the IDE.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSoapRecordReader.java
aielchiieva: Done.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
aielchiieva: Done.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
aielchiieva: Done.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: No, there can be more that 2 items.

Example:
`SELECT DandbCompany.Name, (SELECT LastName, Owner.UserRole.Id FROM Contacts) FROM Account`
Result: set of two maps (first map will contain only one entry, second will contain 4 entries):
```
[
  {Account=TOP_LEVEL}, 
  {Account=TOP_LEVEL, DandbCompany=REFERENCE}, 
  {Account=TOP_LEVEL, Contacts=RELATIONSHIP}, 
  {Account=TOP_LEVEL, Contacts=RELATIONSHIP, Owner=REFERENCE, UserRole=REFERENCE}
]
```
We can't use proposed class structure since we have dependency between relationship and references names order as described in the comment below. I propose to introduce `FeaturedSObjects` class which will include information about relationship and reference SObjects. Top level SObject (i.e. SObject in the `FROM` clause ) is not taken into account since it would be passed separately.
```
  public static class FeaturedSObjects {
    @Nullable
    private String relationship;
    @Nullable
    private List<String> references;
```
For the above query there will be three unique combinations of features SObjects:
`FeaturedSObjects{relationship=null, references=[DandbCompany]}`
`FeaturedSObjects{relationship='Contacts', references=null}`
`FeaturedSObjects{relationship='Contacts', references=[Owner, UserRole]}`

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
aielchiieva: Done.

aielchiieva:
@albertshau addressed second round of code review comments, please review.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
```diff
@@ -400,26 +349,76 @@ public String toString() {
...
+   * there are three unique combinations of features SObjects:
+   * Name -> FeaturedSObjects{relationship=null, references=[DandbCompany]}
+   * FirstName, LastName -> FeaturedSObjects{relationship='Contacts', references=null}
+   * Id -> FeaturedSObjects{relationship='Contacts', references=[Owner, UserRole]}
    */
-  public enum SObjectType {
+  public static class FeaturedSObjects {
+    @Nullable
+    private final String relationship;
+    @Nullable
```
albertshau: It seems like a null is never passed into the constructor. If so, this should be removed. If a null can be passed in, I think it's better that an empty list is set for this rather than allowing this to be nullable.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
```diff
@@ -65,26 +65,60 @@ public SObjectsDescribeResult(PartnerConnection connection, Collection<String> s
...
+
+        /*
+          If featured SObjects have relationship, describe it first.
+          For the query `SELECT Name, (SELECT LastName, Owner.UserRole.Id FROM Contacts) FROM Account`
+          `Contacts` is relationship SObject.
+         */
+        if (obj.hasRelationship()) {
+          String relationshipName = getRelationshipName(obj.getRelationship(), lastDescribe);
+          if (relationshipName == null) {
+            throw new IllegalArgumentException("Unable to find relationship field name: " + name);
```
albertshau: I'm not sure if this error is descriptive enough. For the example above, 'name'=Account and obj.getRelationship()='Contacts' correct? Does this happen if 'Contacts' could not be found? If so a more detailed error message is needed.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
```diff
@@ -65,26 +65,60 @@ public SObjectsDescribeResult(PartnerConnection connection, Collection<String> s
...
+        /*
+          If featured SObjects have references, describe describe them as well.
+          For the query `SELECT Name, (SELECT LastName, Owner.UserRole.Id FROM Contacts) FROM Account`
+          `Owner` and `UserRole` are reference SObjects.
+         */
+        if (obj.hasReferences()) {
+          for (String reference : Objects.requireNonNull(obj.getReferences())) {
+            String referenceName = getReferenceName(reference, lastDescribe);
+            if (referenceName == null) {
+              throw new IllegalArgumentException("Unable to find reference field name: " + name);
```
albertshau: similar comment here

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
```diff
@@ -65,26 +65,60 @@ public SObjectsDescribeResult(PartnerConnection connection, Collection<String> s
...
+            }
+            lastDescribe = describe(connection, referenceName, cache);
+          }
+        }
+
+        // store final describe result to be used during schema creation
+        addSObjectDescribe(obj.getFullName(name), lastDescribe.getFields());
+      }
+    } catch (ConnectionException e) {
+      throw new RuntimeException(e);
```
albertshau: In general, we want to avoid catching a checked exception and throwing a runtime exception. This type of pattern makes it very difficult to do proper error handling, especially if it exists in multiple layers in the code. The caller might not be able to handle it in any meaningful way and may need to propagate it up, but that should be a decision that the caller makes and not this class.

In general we would also like to avoid constructors that throw exceptions or do work, such as making remote calls. A better pattern is to have a constructor that just takes the objectToFieldMap, and turn this into a static method that returns a SObjectsDescribeResult and throws a ConnectionException.

albertshau (commented):
Thanks for creating that new class and adding more comments, it makes the logic much easier to follow.

A couple minor comments left.

src/main/java/io/cdap/plugin/salesforce/SObjectDescriptor.java
aielchiieva: Removed and ensured we always get empty list if null is passed.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
aielchiieva: Updated error message.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
aielchiieva: Updated error message.

src/main/java/io/cdap/plugin/salesforce/SObjectsDescribeResult.java
aielchiieva: Replaced with static methods calls which throws ConnectionException.

aielchiieva:
@albertshau addressed code review comments, please review.

albertshau (approved):
lgtm. Please squash the commits and I'll merge

aielchiieva:
@albertshau done.

PR #35, opened 2019-06-03
CDAP-15313 Salesforce batch sink integration tests

aonischuk:
1. Run test ETL with mock source and sfdc sink
2. Form a SOQL query for every structured record added, which checks every field value. And check whether the query returns a result.
3. Remove created sObjects from Salesforce in `@After`

Also I have added a small refactoring change of working with logical types in this PR.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
@@ -17,10 +17,13 @@ package io.cdap.plugin.salesforce.plugin.sink.batch;
 
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
-import io.cdap.plugin.salesforce.SalesforceTransformUtil;
 
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Transforms a {@link StructuredRecord} to a {@link CSVRecord}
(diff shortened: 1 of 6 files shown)


src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSinkETLTest.java
```diff
@@ -0,0 +1,252 @@
...
+    );
+
+    // Exception below is expected:
+    // "java.lang.IllegalStateException: Expected 200 OK, got 400 Bad Request. Error: Failed to deploy app"
+    // This indicates that validation (not actual execution) failed.
+    deployPipeline(sObject, schema);
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void testNonExistingFieldsValidation() throws Exception {
```
albertshau: It's best not to test validation logic by deploying a pipeline. There are more isolated (and less expensive) ways to test validation. In the past, we've had tests like these that have "passed" because the app failed to deploy, but it happened to fail for a different reason than what the test was expecting. 

For example, if there is a validate method in the conf, an instance of the conf can be created and the validate method called directly. This also allows more granular testing in case different types of exceptions are thrown during validation.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
```diff
@@ -32,61 +29,16 @@
...
 
       fieldNames.add(fieldName);
       values.add(value);
     }
 
-    CSVRecord csvRecord = new CSVRecord(fieldNames, values);
-    return csvRecord;
-  }
-
-  private static Object convert(Object value, Schema.Field field) {
```
albertshau, line 49: I think this method belong in this class more than in some util class. Let's make it public (as well as changing the return type to String).

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSinkETLTest.java
aonischuk: Done. However this still has to remain in integration tests (not unit tests). Since validation process needs connection to Salesforce instance.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/StructuredRecordToCSVRecordTransformer.java
aonischuk: done

aonischuk:
@albertshau addressed your comments. Can we merge at this point?

albertshau (approved):
lgtm

PR #36, opened 2019-06-05
update dependency versions

albertshau:
Use non-snapshot versions of CDAP dependencies and update jackson
to a newer version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,13 +30,13 @@
     <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
-    <cdap.version>6.0.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.0.0</cdap.version>
     <junit.version>4.12</junit.version>
     <commons.io.version>2.6</commons.io.version>
     <commons.lang.version>2.6</commons.lang.version>
     <hadoop.version>2.3.0</hadoop.version>
     <spark2.version>2.1.3</spark2.version>
-    <hydrator.version>2.3.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.2.0</hydrator.version>
     <commons.version>3.8.1</commons.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -30,21 +30,21 @@
...
+    <hydrator.version>2.2.0</hydrator.version>
     <commons.version>3.8.1</commons.version>
     <salesforce.api.version>45.0.0</salesforce.api.version>
     <cometd.java.client.version>4.0.0</cometd.java.client.version>
     <antlr.version>4.7.2</antlr.version>
     <mockito.version>1.10.19</mockito.version>
     <commons.csv.version>1.6</commons.csv.version>
     <jackson.version>1.9.13</jackson.version>
-    <jackson2.version>2.6.7</jackson2.version>
+    <jackson2.version>2.9.9</jackson2.version>
```
CuriousVini, line 47: Why is this change needed?

pom.xml
```diff
@@ -30,21 +30,21 @@
     <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
-    <cdap.version>6.0.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.0.0</cdap.version>
```
CuriousVini, line 33: shouldn't we rely on cdap 6.1.0-SNAPSHOT on develop similar to https://github.com/data-integrations/wrangler/blob/develop/pom.xml#L86?

pom.xml
albertshau: 2.6.7 has a security bug

pom.xml
albertshau: This plugin declares that it is compatible with 6.x versions of the data-pipeline app, which means this needs to be 6.0.0. If it were to use a new API in 6.1.0-SNAPSHOT and then were to be deployed on a 6.0.x version of CDAP, it would fail.

In general, I think we should change our process for plugins such that we don't update the dependent cdap version unless they need to use some new API, so that they're compatible with as many versions of CDAP as possible.

CuriousVini (approved):

PR #37, opened 2019-06-06
CDAP-15455. Support Upsert and Update operations for SFDC Sink

aonischuk:
Without the change only insert is supported.

_How new operations work in Salesforce:_
**Upsert** - Salesforce requires user to provide an external id field name. This field is used as basis for upsert (Salesforce will decide if objects are the same using it).

Id field can be used for that (which is present for all Salesforce sObjects). Also user can create a custom field and checkmark it as "External Id" via Salesforce UI.

**Update** - Specifying an external id field is not supported. For updates Salesforce will always use 'Id' field as basis.

### DIFF
--- a/docs/Salesforce-batchsink.md
+++ b/docs/Salesforce-batchsink.md
@@ -27,10 +27,19 @@ A Salesforce connected application must be created in order to get a client secr
 
 **SObject Name:** Salesforce object name to insert records into.
 
-**Max records per batch:** Maximum number of records to include in a batch when writing to Salesforce.
+**Operation:** Operation used for writing data into Salesforce.<br>
+Insert - adds records.<br>
+Upsert - upserts the records. Salesforce will decide if sObjects 
+are the same using external id field.<br>
+Update - updates existing records based on Id field.
+
+**Upsert External ID Field:** External id field name. It is used only if operation is upsert.
+The field specified can be either 'Id' or any customly created field, which has external id attribute set.
+
+**Max Records Per Batch:** Maximum number of records to include in a batch when writing to Salesforce.
 This value cannot be greater than 10,000.
 
-**Max bytes per batch:** Maximum size in bytes of a batch of records when writing to Salesforce.
+**Max Bytes Per Batch:** Maximum size in bytes of a batch of records when writing to Salesforce.
 This value cannot be greater than 10,000,000.
 
 **Error Handling:** Strategy used to handle erroneous records.<br>
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
@@ -72,12 +72,17 @@ public final class SalesforceBulkUtil {
    * @throws AsyncApiException if there is an issue creating the job
    */
   public static JobInfo createJob(BulkConnection bulkConnection,
-                                  String sObject, OperationEnum operationEnum) throws AsyncApiException {
+                                  String sObject, OperationEnum operationEnum,
+                                  String externalIdField) throws AsyncApiException {
     JobInfo job = new JobInfo();
     job.setObject(sObject);
     job.setOperation(operationEnum);
     job.setConcurrencyMode(ConcurrencyMode.Parallel);
     job.setContentType(ContentType.CSV);
+    if (externalIdField != null) {
+      job.setExternalIdFieldName(externalIdField);
+    }
+
     job = bulkConnection.createJob(job);
     Preconditions.checkState(job.getId() != null, "Couldn't get job ID. There was a problem in creating the " +
       "batch job");
@@ -112,7 +117,7 @@ public final class SalesforceBulkUtil {
     throws AsyncApiException, IOException {
 
     SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
-    JobInfo job = createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query);
+    JobInfo job = createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query, null);
 
     try (ByteArrayInputStream bout = new ByteArrayInputStream(query.getBytes())) {
       bulkConnection.createBatchFromStream(job, bout);
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
@@ -15,6 +15,8 @@
  */
 package io.cdap.plugin.salesforce.plugin.sink.batch;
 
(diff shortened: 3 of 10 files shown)


docs/Salesforce-batchsink.md
```diff
@@ -27,6 +27,15 @@ A Salesforce connected application must be created in order to get a client secr
 
 **SObject Name:** Salesforce object name to insert records into.
 
+**Operation:** Operation used for sinking data into Salesforce.<br>
```
albertshau: sinking -> writing

docs/Salesforce-batchsink.md
```diff
@@ -27,6 +27,15 @@ A Salesforce connected application must be created in order to get a client secr
...
+Insert - adds records.<br>
+Upsert - upserts the records. Salesforce will decide if sObjects 
+are the same using external id field.<br>
+Update - updates existing records based on Id field.
+
+**Upsert external id field:** External id field name. It is used only if operation is upsert.
```
albertshau: By convention the labels are all capitalized. So 'Upsert External ID Field'. 

I missed this in the previous PR, but the same should apply to 'Max Records Per Batch' and 'Max Bytes Per Batch'.

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -112,7 +117,7 @@ public static void closeJob(BulkConnection bulkConnection, String jobId) throws
     throws AsyncApiException, IOException {
 
     SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
-    JobInfo job = createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query);
+    JobInfo job = createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query,  null);
```
albertshau: extra space before 'null'

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -158,21 +207,58 @@ private void validateSchema(Schema schema) {
...
+          SalesforceSinkConfig.PROPERTY_EXTERNAL_ID_FIELD);
+      } else if (!externalIdField.isExternalId() && !externalIdField.getName().equals(SALESFORCE_ID_FIELD)) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' is not configured as external id in Salesforce", externalIdFieldName),
+          SalesforceSinkConfig.PROPERTY_EXTERNAL_ID_FIELD);
+      }
+    } else {
+      if (!Strings.isNullOrEmpty(getExternalIdField())) {
+        throw new InvalidConfigPropertyException(
+          String.format("External id field must be unset for operation='%s'", operation),
```
albertshau: nit: must be unset -> must not be set

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -158,21 +207,58 @@ private void validateSchema(Schema schema) {
...
+        externalIdFieldName = SALESFORCE_ID_FIELD;
+        break;
+      default:
+        throw new InvalidConfigPropertyException("Unsupported value for operation: " + operation,
+                                                 SalesforceSinkConfig.PROPERTY_OPERATION);
+    }
+
+    if (operation == OperationEnum.upsert) {
+      Field externalIdField = describeResult.getField(sObject, externalIdFieldName);
+      if (externalIdField == null) {
```
albertshau, line 243: should we also check the type of the external id field?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -182,20 +268,26 @@ private void validateSchema(Schema schema) {
...
     }
     return creatableSObjectFields;
   }
+
+  private SObjectsDescribeResult getSObjectDescribeResult() {
+    AuthenticatorCredentials credentials = this.getAuthenticatorCredentials();
+    try {
+      PartnerConnection partnerConnection = new PartnerConnection(Authenticator.createConnectorConfig(credentials));
+      SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromName(this.getSObject(),
+                                                     this.getAuthenticatorCredentials());
```
albertshau: nit: indentation

widgets/Salesforce-batchsink.json
```diff
@@ -54,6 +54,24 @@
           "label": "SObject Name",
           "name": "sObject"
         },
+        {
+          "widget-type": "select",
```
albertshau: Since there are only 3 possibilities, I think it's better to use the radio-group widget with "inline" layout. That way the user sees all 3 possibilities without needing to click the dropdown.

https://github.com/cdapio/hydrator-plugins/blob/03f41c467c2e24e87a805afc853e8d2e7bb37bc0/core-plugins/widgets/FTP-batchsource.json#L38

albertshau (commented):
minor comments

docs/Salesforce-batchsink.md
aonischuk: done

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
aonischuk: done

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
aonischuk: Actually I just realized that we don't check input schema for types at all. I have added a general check for all the types. This will include checking externalIdField as well.

ps. This is implemented in "private void validateInputSchema(Schema schema)" method.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
aonischuk: done

widgets/Salesforce-batchsink.json
aonischuk: done

aonischuk:
@albertshau fixed you comments. Can you please merge if it looks fine?

albertshau (approved):
lgtm

PR #39, opened 2019-10-23
[CDAP-15787] Add failure collector apis

CuriousVini:

PR #40, opened 2019-11-01
PLUGIN-87. Salesforce streaming plugin validate button is broken

aonischuk:
This is caused by the fact that new validation framework does not load spark classes during validation as was mentioned to me by Vinisha Shah in a different PR.

The fix involves moving Spark related stuff to a separate package.


aonischuk:
@CuriousVini @albertshau can anybody review/merge please.
This is a pretty simple fix. Just moved methods from one class to another.

CuriousVini:
I already have a PR to solve this issue
https://github.com/data-integrations/salesforce/pull/39

PR #43, opened 2019-11-08
Clarified that the Salesforce password needs to include the User's security token.

ejangi:
This tripped me up while trying to run CDAP locally with Docker. Thought it might be useful to mention in the docs.


ejangi:
@googlebot I signed it!

PR #65, opened 2021-03-12
BugFix/PLUGIN-561 Implement PK Chunking

ardian4:
Add option to enable Primary Key Chunking for Source Plugin

JiraTicket: https://cdap.atlassian.net/browse/PLUGIN-561

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.2.0</cdap.version>
+    <cdap.version>6.4.0-SNAPSHOT</cdap.version>
     <junit.version>4.12</junit.version>
     <commons.io.version>2.6</commons.io.version>
     <commons.lang.version>2.6</commons.lang.version>
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.SequenceInputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -109,21 +110,59 @@ public final class SalesforceBulkUtil {
    *
    * @param bulkConnection bulk connection instance
    * @param query a SOQL query
+   * @param enablePKChunk enable PK Chunk
    * @return an array of batches
    * @throws AsyncApiException  if there is an issue creating the job
    * @throws IOException failed to close the query
    */
-  public static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query)
+  public static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query, boolean enablePKChunk)
     throws AsyncApiException, IOException {
 
     SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
     JobInfo job = createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query, null);
-
+    BatchInfo batchInfo;
     try (ByteArrayInputStream bout = new ByteArrayInputStream(query.getBytes())) {
-      bulkConnection.createBatchFromStream(job, bout);
+      batchInfo = bulkConnection.createBatchFromStream(job, bout);
     }
+    return enablePKChunk ? waitForBatchChunks(bulkConnection, job.getId(), batchInfo.getId()) :
+      bulkConnection.getBatchInfoList(job.getId()).getBatchInfo();
+  }
 
-    return bulkConnection.getBatchInfoList(job.getId()).getBatchInfo();
+  /** When PK Chunk is enabled, wait for state of initial batch to be NotProcessed, in this case Salesforce API will
+   * decide how many batches will be created
+   * @param bulkConnection bulk connection instance
+   * @param jobId a job id
+   * @param initialBatchId a batch id
+   * @return Array with Batches created by Salesforce API
+   *
+   * @throws AsyncApiException if there is an issue creating the job
+   */
+  private static BatchInfo[] waitForBatchChunks(BulkConnection bulkConnection, String jobId, String initialBatchId)
+    throws AsyncApiException {
+    BatchInfo initialBatchInfo = null;
+    for (int i = 0; i < GET_BATCH_RESULTS_TRIES; i++) {
+      //check if the job is aborted
+      if (bulkConnection.getJobStatus(jobId).getState() == JobStateEnum.Aborted) {
+        LOG.info(String.format("Job with Id: '%s' is aborted", jobId));
+        return new BatchInfo[0];
+      }
+      initialBatchInfo = bulkConnection.getBatchInfo(jobId, initialBatchId);
+
+      if (initialBatchInfo.getState() == BatchStateEnum.NotProcessed) {
+        BatchInfo[] result = bulkConnection.getBatchInfoList(jobId).getBatchInfo();
+        return Arrays.stream(result).filter(batchInfo -> batchInfo.getState() != BatchStateEnum.NotProcessed)
+          .toArray(BatchInfo[]::new);
+      } else if (initialBatchInfo.getState() == BatchStateEnum.Failed) {
+        throw new BulkAPIBatchException("Batch failed", initialBatchInfo);
+      } else {
+        try {
+          Thread.sleep(GET_BATCH_RESULTS_SLEEP_MS);
+        } catch (InterruptedException e) {
+          throw new RuntimeException("Job is aborted", e);
+        }
+      }
+    }
+    throw new BulkAPIBatchException("Timeout waiting for batch results", initialBatchInfo);
   }
 
   /**
--- a/src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
+++ b/src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
@@ -304,4 +304,28 @@ public class SalesforceQueryVisitor extends SOQLBaseVisitor<SObjectDescriptor> {
       return Boolean.FALSE;
     }
   }
+
+  /**
+   * Visits query fields that are restricted by Bulk API with PKChunk Enable.
+   * Returns true if there is condition clause other than 'WHERE', false otherwise.
+   * For example: SELECT Name FROM Opportunity LIMIT 10.
+   */
+  public static class RestrictedPKQueryVisitor extends SOQLBaseVisitor<Boolean> {
+
+    @Override
+    public Boolean visitStatement(SOQLParser.StatementContext ctx) {
+      SOQLParser.FromStatementContext fromStatementContext = ctx.fromStatement();
+      if (
+        fromStatementContext.WITH() != null ||
+          fromStatementContext.GROUP() != null ||
+          fromStatementContext.ORDER() != null ||
+          fromStatementContext.LIMIT() != null ||
+          fromStatementContext.OFFSET() != null ||
+          fromStatementContext.FOR() != null) {
+        return true;
+      }
+
+      return ctx.fieldList().accept(new RestrictedFieldVisitor());
+    }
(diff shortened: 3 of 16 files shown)


flakrimjusufi (commented):
Why do we have to change the version of CDAP in pom.xml?

pom.xml
```diff
@@ -31,7 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.2.0</cdap.version>
+    <cdap.version>6.4.0-SNAPSHOT</cdap.version>
```
flakrimjusufi, line 34: Is this change mandatory? Why we have to change CDAP version in pom.xml?

ardian4: > Why do we have to change the version of CDAP in pom.xml?

Plugin has ETL Test, but tests are ignored if there are no Salesforce Account Credentials set in environment variable.  https://github.com/data-integrations/salesforce/blob/f3d38b2470f399a1869e7ad3129796b0dcadf73e/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceETLTest.java#L47-L60

With cdap.version 6.2.0 when etl test are executed we got this issue:

`2021-03-16 12:21:22,777 - WARN  [main:i.c.c.i.a.r.a.ArtifactInspector@252] - Plugin configuration type not supported. Plugin ignored. class io.cdap.plugin.salesforce.plugin.source.batch.SalesforceBatchSource
io.cdap.cdap.api.data.schema.UnsupportedTypeException: Only primitive and String types are supported`

which is related with:

`Plugins use the new macro substitution that supports object, which is introduced in 6.4.0.`

flakrimjusufi: @rmstar  

Prashant, can you give us a hand on this matter? 

Is changing the cdap version in pom.xml the right approach to handle this scenario?

rmstar: Yes, if you have a dependency on an API that was added in 6.4.0 (going to be released in a couple of weeks) then you need to update cdap version in pom.xml.
Once 6.4.0 is released I would suggest changing cdap version from 6.4.0-SNAPSHOT to 6.4.0. Prefer not to depend on snapshot version.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -202,4 +220,49 @@ private void validateCompoundFields(String sObjectName, List<String> fieldNames,
...
+    if (getChunkSize() > SalesforceSourceConstants.MAX_PK_CHUNK_SIZE) {
+      collector.addFailure(
+        String.format("Chunk Size '%d' is bigger that allowed size '%d'.", getChunkSize(),
+                      SalesforceSourceConstants.MAX_PK_CHUNK_SIZE),
+        String.format("Allowed values are between the range of '%d' and '%d'.",
+                      SalesforceSourceConstants.MIN_PK_CHUNK_SIZE, SalesforceSourceConstants.MAX_PK_CHUNK_SIZE))
+        .withConfigProperty(SalesforceSourceConstants.PROPERTY_CHUNK_SIZE_NAME);
+    } else if (getChunkSize() < SalesforceSourceConstants.MIN_PK_CHUNK_SIZE) {
+      collector.addFailure(
+        String.format("Chunk Size %d is lower that allowed size '%d'.", getChunkSize(),
```
rmstar: lower than minimum allowed size.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -202,4 +220,49 @@ private void validateCompoundFields(String sObjectName, List<String> fieldNames,
...
+            String.format("SOQL Query contains restricted clauses when PK Chunk is Enabled. Unsupported query: '%s'.",
+                          query),
+            "Set Enable PK Chunk to false, because 'WHERE' is the only supported conditions clause.")
+            .withConfigProperty(SalesforceSourceConstants.PROPERTY_QUERY);
+        }
+    }
+
+    if (getChunkSize() > SalesforceSourceConstants.MAX_PK_CHUNK_SIZE) {
+      collector.addFailure(
+        String.format("Chunk Size '%d' is bigger that allowed size '%d'.", getChunkSize(),
```
rmstar: bigger than maximum allowed size

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -29,14 +29,25 @@
...
+  public static final String CONFIG_PK_CHUNK_ENABLE = "mapred.salesforce.input.pk.chunk";
+  public static final String CONFIG_CHUNK_SIZE = "mapred.salesforce.input.schemas.chunk.size";
+  public static final String HEADER_ENABLE_PK_CHUNK = "Sforce-Enable-PKChunking";
+  public static final String HEADER_VALUE_PK_CHUNK = "chunkSize=%d";
+
   public static final String CONFIG_SOBJECT_NAME_FIELD = "mapred.salesforce.input.sObjectNameField";
 
   public static final int WIDE_QUERY_MAX_BATCH_COUNT = 2000;
-
+  public static final int MAX_PK_CHUNK_SIZE = 250000;
```
rmstar, line 55: How did we decide that max PK chunk size = 250000? If it's a limit set by salesforce, please add a link to their documentation.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -304,4 +304,28 @@ protected Boolean defaultResult() {
...
+      SOQLParser.FromStatementContext fromStatementContext = ctx.fromStatement();
+      if (
+        fromStatementContext.WITH() != null ||
+          fromStatementContext.GROUP() != null ||
+          fromStatementContext.ORDER() != null ||
+          fromStatementContext.LIMIT() != null ||
+          fromStatementContext.OFFSET() != null ||
+          fromStatementContext.FOR() != null) {
+        return true;
+      }
```
rmstar, line 326: should we also validate the objects name (In the documentation you have a list of supported objects)?

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
```diff
@@ -109,21 +110,58 @@ public static void closeJob(BulkConnection bulkConnection, String jobId) throws
...
+   *
+   * @throws AsyncApiException if there is an issue creating the job
+   */
+  private static BatchInfo[] waitForBatchChunks(BulkConnection bulkConnection, String jobId, String initialBatchId)
+    throws AsyncApiException {
+    BatchInfo initialBatchInfo = null;
+    for (int i = 0; i < GET_BATCH_RESULTS_TRIES; i++) {
+      //check if the job is aborted
+      if (bulkConnection.getJobStatus(jobId).getState() == JobStateEnum.Aborted) {
+        throw new RuntimeException(new RuntimeException(String.format("Job with Id: '%s' is aborted", jobId)));
```
rmstar: Why are we throwing a RuntimeException when the the api returns an error? Can we throw a checked exception instead?

pom.xml
rmstar: Can you clarify what change was made in this PR that requires CDAP 6.4.0 dependency? I'd like to avoid bumping up the version of CDAP dependency if possible.

ardian4: There was no specific requirement in the PR that required this dependency change but when we added ETL Tests we faced this issue. 

To verify this, we did the following: 

1. We run the ETL Tests from the develop branch without our changes, and we encountered the same issue. 
2. When we checked out the history from git, the minimum version of CDAP was changed to 6.4.0 but we think that after this change the ETL tests were skipped because you need to provide Salesforce account credentials. 

You can verify this by checking this out:
https://github.com/data-integrations/salesforce/blob/08b5980eec121c23a4529a5d5bcf805c5003776c/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceETLTest.java#L81-L89

Please let us know if there's any other way to run only the ETL Tests in CDAP 6.4.0.

rmstar: Chatted with @chtyim about this - on develop branch there is a dependency on CDAP 6.4.0 introduced by recent oauth changes. 
However, on release/1.3 branch we don't need to bump up the dependency.
So let's do the following:
1. Merge this PR with CDAP depedency version set to 6.4.0.
2. When you backport this PR to release/1.3 branch, set cdap.version = 6.2.0.

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
ardian4: We changed logic here, we return zero splits and log info message.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
ardian4: Updated

src/test/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfigTest.java
```diff
@@ -185,4 +191,91 @@ public void testGetOffsetException() {
...
+
+  private void testValidPKChunkConfiguration(SalesforceSourceConfig config) {
+    MockFailureCollector collector = new MockFailureCollector();
+    SalesforceSourceConfig mock = Mockito.spy(config);
+    Mockito.when(mock.canAttemptToEstablishConnection()).thenReturn(false);
+    try {
+      mock.validate(collector);
+      Assert.assertEquals(0, collector.getValidationFailures().size());
+    } catch (ValidationException e) {
+      Assert.assertEquals(0, e.getFailures().size());
```
rmstar: We don't expect an exception here, right? So don't catch exception here?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -29,14 +32,95 @@
...
+  public static final String CONFIG_PK_CHUNK_ENABLE = "mapred.salesforce.input.pk.chunk";
+  public static final String CONFIG_CHUNK_SIZE = "mapred.salesforce.input.schemas.chunk.size";
+  public static final String HEADER_ENABLE_PK_CHUNK = "Sforce-Enable-PKChunking";
+  public static final String HEADER_VALUE_PK_CHUNK = "chunkSize=%d";
+
   public static final String CONFIG_SOBJECT_NAME_FIELD = "mapred.salesforce.input.sObjectNameField";
 
   public static final int WIDE_QUERY_MAX_BATCH_COUNT = 2000;
-
+  //https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/
```
rmstar: nit: space after //

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -29,14 +32,95 @@
...
+  //async_api_headers_enable_pk_chunking.htm
+  public static final int MAX_PK_CHUNK_SIZE = 250000;
+  public static final int DEFAULT_PK_CHUNK_SIZE = 100000;
+  public static final int MIN_PK_CHUNK_SIZE = 1;
+  //https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/
```
rmstar: nit: space after //

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
rmstar: Is this not considered to be an error condition where we need to throw an exception?

rmstar (approved):
Added some minor comments.  Please squash commits after addressing them. Thanks!

src/main/java/io/cdap/plugin/salesforce/SalesforceBulkUtil.java
ardian4: Job is Aborted only when user abort it from Salesforce site, so we thought this is not exception.

PR #61, opened 2021-02-17
(CDAP-17611) Added OAuth function support

chtyim:

PR #66, opened 2021-03-16
Bugfix PLUGIN-582

ardian4:
fixed: issue with Time type conversion
added: unit test for Time type conversion
added: integration test for Date, DateTime, Time

JiraTicket:  https://cdap.atlassian.net/browse/PLUGIN-582

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.2.0</cdap.version>
+    <cdap.version>6.4.0-SNAPSHOT</cdap.version>
     <junit.version>4.12</junit.version>
     <commons.io.version>2.6</commons.io.version>
     <commons.lang.version>2.6</commons.lang.version>
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceTransformUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceTransformUtil.java
@@ -20,6 +20,7 @@ import io.cdap.cdap.api.data.schema.Schema;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -43,7 +44,7 @@ public class SalesforceTransformUtil {
       case TIMESTAMP_MICROS:
         return TimeUnit.MILLISECONDS.toMicros(Instant.parse(value).toEpochMilli());
       case TIME_MICROS:
-        return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(value).toNanoOfDay());
+        return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(value, DateTimeFormatter.ISO_TIME).toNanoOfDay());
       default:
         throw new IllegalArgumentException(
           String.format("Field '%s' is of unsupported type '%s'", fieldName, logicalType.getToken()));
--- a/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSourceETLTest.java
+++ b/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSourceETLTest.java
@@ -187,6 +187,24 @@ public abstract class BaseSalesforceBatchSourceETLTest extends BaseSalesforceETL
     return customField;
   }
 
+  protected CustomField createDateCustomField(String fullName) {
+    CustomField customField = createCustomField(fullName);
+    customField.setType(FieldType.Date);
+    return customField;
+  }
+
+  protected CustomField createDateTimeCustomField(String fullName) {
+    CustomField customField = createCustomField(fullName);
+    customField.setType(FieldType.DateTime);
+    return customField;
+  }
+
+  protected CustomField createTimeCustomField(String fullName) {
+    CustomField customField = createCustomField(fullName);
+    customField.setType(FieldType.Time);
+    return customField;
+  }
+
   protected CustomField createLocationCustomField(String fullName) {
     CustomField customField = createCustomField(fullName);
     customField.setType(FieldType.Location);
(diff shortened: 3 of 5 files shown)


src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSourceETLTest.java
```diff
@@ -225,7 +243,7 @@ private static MetadataConnection createMetadataConnection() throws ConnectionEx
     loginConfig.setAuthEndpoint(METADATA_LOGIN_URL);
     loginConfig.setServiceEndpoint(METADATA_LOGIN_URL);
     loginConfig.setManualLogin(true);
-    LoginResult loginResult = new PartnerConnection(loginConfig).login(USERNAME, PASSWORD);
+    LoginResult loginResult = new PartnerConnection(loginConfig).login(USERNAME, PASSWORD + SECURITY_TOKEN);
```
flakrimjusufi, line 246: Why we are concatenating PASSWORD and SECURITY_TOKEN in here?

pom.xml
```diff
@@ -31,7 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.2.0</cdap.version>
+    <cdap.version>6.4.0-SNAPSHOT</cdap.version>
```
flakrimjusufi, line 34: Any specific reason why we are changing the CDAP version in here?

src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSourceETLTest.java
ardian4: In the ticket https://cdap.atlassian.net/browse/PLUGIN-420 was required to append security token.

pom.xml
ardian4: Plugin has ETL Test, but tests are ignored if there are no Salesforce Account Credentials set in environment variable.  https://github.com/data-integrations/salesforce/blob/f3d38b2470f399a1869e7ad3129796b0dcadf73e/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceETLTest.java#L47-L60

With cdap.version 6.2.0 when etl test are executed we got this issue:

`2021-03-16 12:21:22,777 - WARN  [main:i.c.c.i.a.r.a.ArtifactInspector@252] - Plugin configuration type not supported. Plugin ignored. class io.cdap.plugin.salesforce.plugin.source.batch.SalesforceBatchSource
io.cdap.cdap.api.data.schema.UnsupportedTypeException: Only primitive and String types are supported`

which is related with:

`Plugins use the new macro substitution that supports object, which is introduced in 6.4.0.`

CuriousVini: I am still not clear why this change is needed? Isnt time type supported by cdap schema for 6.2 version?

src/main/java/io/cdap/plugin/salesforce/SalesforceTransformUtil.java
```diff
@@ -43,7 +44,7 @@ public static Object transformLogicalType(String fieldName, Schema.LogicalType l
       case TIMESTAMP_MICROS:
         return TimeUnit.MILLISECONDS.toMicros(Instant.parse(value).toEpochMilli());
       case TIME_MICROS:
-        return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(value).toNanoOfDay());
+        return TimeUnit.NANOSECONDS.toMicros(LocalTime.parse(value, DateTimeFormatter.ISO_TIME).toNanoOfDay());
```
CuriousVini, line 47: Is this util used for both source and sink?

ardian4: No, this is just for source. Does the JIRA ticket also include Sink plugin as well?

pom.xml
ardian4: If you run the ETL Tests that are actually in the repo: 

https://github.com/data-integrations/salesforce/blob/08b5980eec121c23a4529a5d5bcf805c5003776c/src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceETLTest.java#L81-L89 you must specify Salesforce credentials here. 

Thus the ETL pipeline will fail with:

`2021-03-16 12:21:22,777 - WARN [main:i.c.c.i.a.r.a.ArtifactInspector@252] - Plugin configuration type not supported. Plugin ignored. class io.cdap.plugin.salesforce.plugin.source.batch.SalesforceBatchSource io.cdap.cdap.api.data.schema.UnsupportedTypeException: Only primitive and String types are supported`

By default, there are no credentials provided for Salesforce account and for this reason all the ETL Tests were skipped.

Please let us know if there's any other way we can run the ETL Tests in CDAP 6.4.0 without bumping the version.

src/main/java/io/cdap/plugin/salesforce/SalesforceTransformUtil.java
CuriousVini: so `SalesforceTransformUtil.java` is not used in batch sink? Is that a correct statement?

pom.xml
CuriousVini: If you run the same test in 6.2 what happens then? the credentials are related to plugin. I am not sure why it matters what CDAP version we are using.

ardian4: When we run the same test in 6.2 we got this issue:

`2021-03-16 12:21:22,777 - WARN [main:i.c.c.i.a.r.a.ArtifactInspector@252] - Plugin configuration type not supported. Plugin ignored. class io.cdap.plugin.salesforce.plugin.source.batch.SalesforceBatchSource io.cdap.cdap.api.data.schema.UnsupportedTypeException: Only primitive and String types are supported`

we had same issue with PLUGIN-561 here is the discussion and solution https://github.com/data-integrations/salesforce/pull/65#discussion_r599082473

src/main/java/io/cdap/plugin/salesforce/SalesforceTransformUtil.java
ardian4: SalesforceTransformUtil.java is not used in batch sink.

pom.xml
CuriousVini: Is this only needed for develop? This indicates plugin code from this branch will not work on cdap versions lower than 6.4?

MEseifan: Maybe I can help clarify, nothing in this PR needs 6.4.0. The previous changes added in #61 added the 6.4.0 requirement. This PR is just updating the dependency so that tests can be run in the develop branch.

This code can be directly cherry-picked into the release/1.3 branch and we can leave the CDAP version as 6.2.0 there. But any future releases cut from develop will only work on 6.4.0+ due to the OAuth changes

CuriousVini: so the code for develop and release branch is diverged at this point. In that case, we should make sure the parent artifacts have corresponding cdap version so that plugins will not show up in any cdap versions lower than 6.4

MEseifan: Yes, the parents on develop is already set for 6.4.0-SNAPSHOT, this was done as part of #61

CuriousVini: sounds good

CuriousVini (approved):

PR #67, opened 2021-03-16
PLUGIN-574: Updated documentation of Salesforce plugins with entities that are not supported by bulk API

flakrimjusufi:
Salesforce entities not supported by plugins
JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-574

### DIFF
--- a/docs/Salesforce-batchsink.md
+++ b/docs/Salesforce-batchsink.md
@@ -30,6 +30,46 @@ A Salesforce connected application must be created in order to get a client secr
 
 **SObject Name:** Salesforce object name to insert records into.
 
+There are also **sObjects** that are not supported in the Bulk API of Salesforce.
+When a job is created using an object that is not supported in the Bulk API, "_Entity is not supported by the Bulk API_" is thrown.
+These objects are also not supported by _Einstein Analytics_ as it also uses Bulk API for querying data.
+
+Below is a non-comprehensive list of **sObjects** that are not currently available in the Bulk API:
+- *Feed (e.g. AccountFeed, AssetFeed, ...)
+- *Share (e.g. AccountBrandShare, ChannelProgramLevelShare, ...)
+- *History (e.g. AccountHistory, ActivityHistory, ...)
+- *EventRelation (e.g. AcceptedEventRelation, DeclinedEventRelation, ...)
+- AggregateResult
+- AttachedContentDocument
+- CaseStatus
+- CaseTeamMember
+- CaseTeamRole
+- CaseTeamTemplate
+- CaseTeamTemplateMember
+- CaseTeamTemplateRecord
+- CombinedAttachment
+- ContentFolderItem
+- ContractStatus
+- EventWhoRelation
+- FolderedContentDocument
+- KnowledgeArticleViewStat
+- KnowledgeArticleVoteStat
+- LookedUpFromActivity
+- Name
+- NoteAndAttachment
+- OpenActivity
+- OwnedContentDocument
+- PartnerRole
+- RecentlyViewed
+- ServiceAppointmentStatus
+- SolutionStatus
+- TaskPriority
+- TaskStatus
+- TaskWhoRelation
+- UserRecordAccess
+- WorkOrderLineItemStatus
+- WorkOrderStatus
+
 **Operation:** Operation used for writing data into Salesforce.<br>
 Insert - adds records.<br>
 Upsert - upserts the records. Salesforce will decide if sObjects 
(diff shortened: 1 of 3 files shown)


docs/Salesforce-streamingsource.md
```diff
@@ -69,6 +69,46 @@ for the evaluated records only if they match the criteria specified in the WHERE
 Salesforce and generate SOQL query (`select <FIELD_1, FIELD_2, ..., FIELD_N> from ${sObjectName}`).
 Ignored if SOQL query is provided.
 
+There are also **sObjects** that are not supported in the Bulk API of Salesforce.
+When a job is created using an object that is not supported in the Bulk API, "_Entity is not supported by the Bulk API_" is thrown.
```
CuriousVini: when would plugin use build api? How would plugin user know it?

CuriousVini: are we using bulk api in streaming source?

flakrimjusufi: Thanks for catching this. Streaming source does not use Bulk API. I removed it from the documentation of Salesfoce-streamingsource in the last update.

docs/Salesforce-batchsink.md
```diff
@@ -30,6 +30,46 @@ A Salesforce connected application must be created in order to get a client secr
 
 **SObject Name:** Salesforce object name to insert records into.
 
+There are also **sObjects** that are not supported in the Bulk API of Salesforce.
+When a job is created using an object that is not supported in the Bulk API, "_Entity is not supported by the Bulk API_" is thrown.
```
CuriousVini, line 34: do we always use bulk api for source and sink? Are there any cases when we use non-bulk api?

docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -158,3 +158,43 @@ multisink.Activity =
     ]
   }
 ```
+
+There are also **sObjects** that are not supported in the Bulk API of Salesforce.
+When a job is created using an object that is not supported in the Bulk API, "_Entity is not supported by the Bulk API_" is thrown.
```
CuriousVini, line 163: does multi object source also use bulk api?

docs/Salesforce-batchsink.md
flakrimjusufi: There's one case when we don't use bulk API in source and sink. 

According to java doc in **SalesforceInputFormat.java**: 

```
Based on query length sends query to Salesforce to receive array of batch info.
If query is within limit, executes original query. If not, switches to wide object logic,
i.e. generates Id query to retrieve batch info for Ids only that will be used later
to retrieve data using SOAP API.
```

https://github.com/data-integrations/salesforce/blob/48270f0815e85f1f62b7cf73c7acf4a80a6563fe/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java#L111-L133

docs/SalesforceMultiObjects-batchsource.md
flakrimjusufi: SalesForceBatchMultiSource uses SalesforceInputFormatProvider which uses Bulk Api. 

https://github.com/data-integrations/salesforce/blob/48270f0815e85f1f62b7cf73c7acf4a80a6563fe/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchMultiSource.java#L84-L85

docs/Salesforce-batchsink.md
CuriousVini: I see. In that case, Could you please make sure documentation reflects those cases.

docs/SalesforceMultiObjects-batchsource.md
CuriousVini: I see. sounds good

docs/Salesforce-batchsink.md
flakrimjusufi: > I see. In that case, Could you please make sure documentation reflects those cases.

PR updated accordingly.

rmstar (approved):

docs/SalesforceMultiObjects-batchsource.md
```diff
@@ -158,3 +158,43 @@ multisink.Activity =
     ]
   }
 ```
```
rmstar, line 160: Is the `Cases when BULK API is not used` section also applicable to multi object source? If yes, can you please add it here?

flakrimjusufi: Yes, this section is also applicable to multi object source and it's already added in the documentation.

PR #72, opened 2021-03-31
Added support for defining parent object for PK chunking

MEseifan:
The Salesforce [chunking option](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/async_api_headers_enable_pk_chunking.htm) supports an additional "parent" header that can be used to define a parent object that can be used for chunking. This is used when the object being fetched is a history table or a shared object.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
@@ -45,8 +45,8 @@ import javax.annotation.Nullable;
 import static io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSourceConstants.SUPPORTED_OBJECTS_WITH_PK_CHUNK;
 
 /**
- * This class {@link SalesforceSourceConfig} provides all the configuration required for
- * configuring the {@link SalesforceBatchSource} plugin.
+ * This class {@link SalesforceSourceConfig} provides all the configuration required for configuring the {@link
+ * SalesforceBatchSource} plugin.
  */
 public class SalesforceSourceConfig extends SalesforceBaseSourceConfig {
 
@@ -81,6 +81,12 @@ public class SalesforceSourceConfig extends SalesforceBaseSourceConfig {
   @Description("Specify size of chunk. Maximum Size is 250,000. Default Size is 100,000.")
   private Integer chunkSize;
 
+  @Name(SalesforceSourceConstants.PROPERTY_PARENT_NAME)
+  @Macro
+  @Nullable
+  @Description("Parent of the Salesforce Object. This is used to enable chunking for history tables or shared objects.")
+  private String parent;
+
   @VisibleForTesting
   SalesforceSourceConfig(String referenceName,
                          @Nullable String consumerKey,
@@ -97,7 +103,8 @@ public class SalesforceSourceConfig extends SalesforceBaseSourceConfig {
                          @Nullable String schema,
                          @Nullable String securityToken,
                          @Nullable Boolean enablePKChunk,
-                         @Nullable Integer chunkSize) {
+                         @Nullable Integer chunkSize,
+                         @Nullable String parent) {
     super(referenceName, consumerKey, consumerSecret, username, password, loginUrl,
           datetimeAfter, datetimeBefore, duration, offset, securityToken);
     this.query = query;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -273,8 +297,8 @@ private void checkForPKSupportedObject(String sObject, FailureCollector collecto
     if (canAttemptToEstablishConnection()) {
       if (!isCustomObject(sObject, collector)) {
         if (!SUPPORTED_OBJECTS_WITH_PK_CHUNK.contains(sObject)) {
-          collector.addFailure(String.format("SObject: %s is not supported with PKChunk enabled.", sObject),
-                               "Please check documentation for supported Objects.");
+          collector.addFailure(String.format("SObject '%s' is not supported with PKChunk enabled.", sObject),
+                               "Please check documentation for supported Objects. If this is a history or shared table, you may need to specify a SObject Parent in the Advanced section.");
```
rmstar: nit: split long line

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -237,19 +251,29 @@ private void validatePKChunk(FailureCollector collector) {
...
+      }
 
-      String sObject = SalesforceQueryParser.getObjectDescriptorFromQuery(query).getName();
-      checkForPKSupportedObject(sObject, collector);
+      // If a parent object is defined then use that to check for PK support, otherwise use the object from the query
+      if (!containsMacro(SalesforceSourceConstants.PROPERTY_PARENT_NAME) && !Strings.isNullOrEmpty(getParent())) {
+        checkForPKSupportedObject(getParent(), collector);
+      } else {
+        String sObject = SalesforceQueryParser.getObjectDescriptorFromQuery(query).getName();
+        checkForPKSupportedObject(sObject, collector);
```
rmstar, line 266: if parent name is a macro, then we're checking the object from the query instead of the parent object. Is that expected?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -237,19 +251,29 @@ private void validatePKChunk(FailureCollector collector) {
...
+      }
     }
 
+    // If a parent object is defined then use that to check for PK support, otherwise use the object from the config
     if (!containsMacro(SalesforceSourceConstants.PROPERTY_SOBJECT_NAME) && !Strings.isNullOrEmpty(getSObjectName())) {
-      checkForPKSupportedObject(getSObjectName(), collector);
+      if (!containsMacro(SalesforceSourceConstants.PROPERTY_PARENT_NAME) && !Strings.isNullOrEmpty(getParent())) {
+        checkForPKSupportedObject(getParent(), collector);
+      } else {
+        checkForPKSupportedObject(getSObjectName(), collector);
```
rmstar, line 275: same comment here. we're using the object from config if parent name is a macro.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
MEseifan: done

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
MEseifan: if the parent field is a marco then it wont get to this point, the first check in this function exits if any of the properties are macros.

We technically dont need any of the marco checks after the first one but I left it in there for readability

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
MEseifan: same response as above

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
rmstar: I see, thanks

rmstar (approved):

PR #85, opened 2021-06-30
(PLUGIN-711) Close the bulk job when closing record reader

bdmogal:
First commit is the fix.

Second commit is just cleanup common static code that was not really "common"


bdmogal:
After the fix:

image

MEseifan (approved):

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -123,6 +133,11 @@ public void close() throws IOException {
       // this also closes the inputStream
       csvParser.close();
     }
+    try {
+      SalesforceBulkUtil.closeJob(bulkConnection, jobId);
```
avirmani, line 137: Can we just check that the bulkConnection is not null - in case the exception was thrown before or during its instantiation

avirmani (approved):
Just 1 comment.

PR #89, opened 2021-09-13
Salesforce source fixes

rmstar:
- Retry bulkConnection.getBatchInfo() failures
- Log job, batch status when PK chunking is disabled
- Don't keep inputstreams open

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
@@ -20,7 +20,6 @@ import com.sforce.async.AsyncApiException;
 import com.sforce.async.BatchInfo;
 import com.sforce.async.BatchStateEnum;
 import com.sforce.async.BulkConnection;
-import com.sforce.async.QueryResultList;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.salesforce.BulkAPIBatchException;
 import io.cdap.plugin.salesforce.SalesforceBulkUtil;
@@ -41,12 +40,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.SequenceInputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 
 /**
@@ -64,9 +59,22 @@ public class SalesforceBulkRecordReader extends RecordReader<Schema, Map<String,
   private Map<String, ?> value;
   private String jobId;
   private BulkConnection bulkConnection;
+  private String batchId;
+  private String[] resultIds;
+  private int resultIdIndex;
+  private String[] csvHeaders;
 
   public SalesforceBulkRecordReader(Schema schema) {
+    this(schema, null, null, null);
+  }
+
+  @VisibleForTesting
+  SalesforceBulkRecordReader(Schema schema, String jobId, String batchId, String[] resultIds) {
     this.schema = schema;
+    this.resultIdIndex = 0;
+    this.jobId = jobId;
+    this.batchId = batchId;
+    this.resultIds = resultIds;
   }
 
   /**
@@ -83,15 +91,15 @@ public class SalesforceBulkRecordReader extends RecordReader<Schema, Map<String,
 
     SalesforceSplit salesforceSplit = (SalesforceSplit) inputSplit;
     jobId = salesforceSplit.getJobId();
-    String batchId = salesforceSplit.getBatchId();
+    batchId = salesforceSplit.getBatchId();
     LOG.debug("Executing Salesforce Batch Id: '{}' for Job Id: '{}'", batchId, jobId);
 
     Configuration conf = taskAttemptContext.getConfiguration();
     try {
       AuthenticatorCredentials credentials = SalesforceConnectionUtil.getAuthenticatorCredentials(conf);
       bulkConnection = new BulkConnection(Authenticator.createConnectorConfig(credentials));
-      InputStream queryResponseStream = waitForBatchResults(bulkConnection, jobId, batchId);
-      setupParser(queryResponseStream);
+      resultIds = waitForBatchResults(bulkConnection);
+      setupParser();
     } catch (AsyncApiException e) {
       throw new RuntimeException("There was issue communicating with Salesforce", e);
     }
@@ -103,9 +111,27 @@ public class SalesforceBulkRecordReader extends RecordReader<Schema, Map<String,
    * @return returns false if no more data to read
    */
   @Override
-  public boolean nextKeyValue() {
-    if (!parserIterator.hasNext()) {
+  public boolean nextKeyValue() throws IOException {
+    if (parserIterator == null) {
       return false;
+    }
+    while (!parserIterator.hasNext()) {
+      if (resultIdIndex == resultIds.length) {
+        // No more result ids to process.
+        return false;
+      }
+      // Close CSV parser for previous result.
+      if (csvParser != null && !csvParser.isClosed()) {
+        // this also closes the inputStream
+        csvParser.close();
+        csvParser = null;
+      }
+      try {
+        // Parse the next result.
+        setupParser();
+      } catch (AsyncApiException e) {
+        throw new IOException("Failed to query results", e);
+      }
     }
 
     value = parserIterator.next().toMap();
@@ -129,9 +155,10 @@ public class SalesforceBulkRecordReader extends RecordReader<Schema, Map<String,
 
   @Override
   public void close() throws IOException {
-    if (csvParser != null) {
+    if (csvParser != null && !csvParser.isClosed()) {
       // this also closes the inputStream
       csvParser.close();
+      csvParser = null;
     }
     if (bulkConnection != null) {
       try {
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -169,24 +176,23 @@ void setupParser(InputStream queryResponseStream) throws IOException {
...
-        List<InputStream> streams = new ArrayList<>(resultIds.length);
-        for (String resultId : resultIds) {
-          streams.add(bulkConnection.getQueryResultStream(jobId, batchId, resultId));
+      try {
+        info = bulkConnection.getBatchInfo(jobId, batchId);
+      } catch (AsyncApiException e) {
+        if (i == SalesforceSourceConstants.GET_BATCH_RESULTS_TRIES - 1) {
+          throw e;
         }
+        LOG.warn("Failed to get batch info: {}, retry {}", e, i);
```
albertshau: exceptions should be the last thing in the log. Something like:

`LOG.warn("Failed to get info for batch {}. Will retry after some time.", batchId, e);`

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,19 +145,24 @@ public void close() throws IOException {
...
   @VisibleForTesting
-  void setupParser(InputStream queryResponseStream) throws IOException {
+  void setupParser(String jobId, String batchId, String[] resultIds) throws IOException, AsyncApiException {
     CSVFormat csvFormat = CSVFormat.DEFAULT.
       withHeader().
       withQuoteMode(QuoteMode.ALL).
       withAllowMissingColumnNames(false);
 
-    csvParser = CSVParser.parse(queryResponseStream, StandardCharsets.UTF_8, csvFormat);
+    for (String resultId: resultIds) {
```
albertshau: add a space before ':'

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,19 +145,24 @@ public void close() throws IOException {
...
+      InputStream queryResponseStream = bulkConnection.getQueryResultStream(jobId, batchId, resultId);
```
albertshau: This looks like it doesn't fix the problem, as each InputStream still gets opened here.

Instead, keep track of the result id array along with an index into that array. In nextKeyValue(), if the CSVParser has no more elements and there are more results, open the InputStream for the next result. If there are no more results, the reader is done.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -212,7 +220,15 @@ private BulkConnection getBulkConnection(Configuration conf) {
...
         return new BatchInfo[0];
       }
-      initialBatchInfo = bulkConnection.getBatchInfo(jobId, initialBatchId);
+      try {
+        initialBatchInfo = bulkConnection.getBatchInfo(jobId, initialBatchId);
+      } catch (AsyncApiException e) {
+        if (i == SalesforceSourceConstants.GET_BATCH_RESULTS_TRIES - 1) {
+          throw e;
+        }
+        LOG.warn("Failed to get batch info: {}, retry {}", e, i);
```
albertshau: similar comment here, exceptions should be the last element in a log statement, and should not have a corresponding placeholder `{}`.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -90,8 +90,8 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
     try {
       AuthenticatorCredentials credentials = SalesforceConnectionUtil.getAuthenticatorCredentials(conf);
       bulkConnection = new BulkConnection(Authenticator.createConnectorConfig(credentials));
-      InputStream queryResponseStream = waitForBatchResults(bulkConnection, jobId, batchId);
-      setupParser(queryResponseStream);
+      String[] resultIds = waitForBatchResults(bulkConnection, jobId, batchId);
```
snehalakshmisha: Could resultIds be null?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,19 +145,24 @@ public void close() throws IOException {
...
+      }
+      csvParsers.add(csvParser);
     }
 
-    parserIterator = csvParser.iterator();
+    parserIterator = Iterators.concat(csvParsers.stream().map(CSVParser::iterator)
+                                        .collect(Collectors.toList()).iterator());
   }
 
   /**
```
snehalakshmisha, line 201: @return should be updated in this javadoc

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: this code was removed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: thanks, fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: It won't be null. https://github.com/forcedotcom/wsc/blob/7fcabe96ce4071d0f3693c05636b829f113b9496/src/main/java/com/sforce/async/QueryResultList.java#L53

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,50 +170,57 @@ public void close() throws IOException {
...
+  void setupParser() throws IOException, AsyncApiException {
+    if (resultIdIndex < resultIds.length) {
+      InputStream queryResponseStream = bulkConnection.getQueryResultStream(jobId, batchId, resultIds[resultIdIndex]);
+
+      if (resultIdIndex == 0) {
+        CSVFormat csvFormat = CSVFormat.DEFAULT.
+          withHeader().
+          withQuoteMode(QuoteMode.ALL).
+          withAllowMissingColumnNames(false);
+         csvParser = CSVParser.parse(queryResponseStream, StandardCharsets.UTF_8, csvFormat);
```
albertshau: fix indentation

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,50 +170,57 @@ public void close() throws IOException {
...
-    csvParser = CSVParser.parse(queryResponseStream, StandardCharsets.UTF_8, csvFormat);
-
-    if (csvParser.getHeaderMap().isEmpty()) {
-      throw new IllegalStateException("Empty response was received from Salesforce, but csv header was expected.");
+  void setupParser() throws IOException, AsyncApiException {
+    if (resultIdIndex < resultIds.length) {
+      InputStream queryResponseStream = bulkConnection.getQueryResultStream(jobId, batchId, resultIds[resultIdIndex]);
+
+      if (resultIdIndex == 0) {
+        CSVFormat csvFormat = CSVFormat.DEFAULT.
```
albertshau: nit: put the dot before the method call

```
CSVFormat.DEFAULT
  .withHeader()
  .withQuoteMode(...
```

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,50 +170,57 @@ public void close() throws IOException {
...
-      withHeader().
-      withQuoteMode(QuoteMode.ALL).
-      withAllowMissingColumnNames(false);
-
-    csvParser = CSVParser.parse(queryResponseStream, StandardCharsets.UTF_8, csvFormat);
-
-    if (csvParser.getHeaderMap().isEmpty()) {
-      throw new IllegalStateException("Empty response was received from Salesforce, but csv header was expected.");
+  void setupParser() throws IOException, AsyncApiException {
+    if (resultIdIndex < resultIds.length) {
```
albertshau: This should never be called if there are no more results, so it's better to throw an exception in that case instead of have it be a no-op. Also more readable to check the inverse to remove some nesting and a large if block:

```
if (resultIdIndex >= resultIds.length) {
  throw new IllegalArgumentException(...);
}
...
```

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -103,10 +111,28 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
    */
   @Override
-  public boolean nextKeyValue() {
-    if (!parserIterator.hasNext()) {
+  public boolean nextKeyValue() throws IOException {
+    if (parserIterator == null) {
       return false;
     }
+    if (!parserIterator.hasNext()) {
+      if (resultIdIndex == resultIds.length) {
```
albertshau, line 119: If a result can contain empty data, this will result in an error. Similarly, it seems like this would behave incorrectly if the first result contains only a header and no actual data. 

Better to put this in a loop, until the parserIterator does have a next value.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -190,8 +190,16 @@ private BulkConnection getBulkConnection(Configuration conf) {
...
-    return enablePKChunk ? waitForBatchChunks(bulkConnection, job.getId(), batchInfo.getId()) :
-      bulkConnection.getBatchInfoList(job.getId()).getBatchInfo();
+
+    if (enablePKChunk) {
+      return waitForBatchChunks(bulkConnection, job.getId(), batchInfo.getId());
+    }
+    BatchInfo[] batchInfos = bulkConnection.getBatchInfoList(job.getId()).getBatchInfo();
+    LOG.info("Job id {}, status: {}", job.getId(), bulkConnection.getJobStatus(job.getId()).getState());
+    if (batchInfos.length > 0) {
+      LOG.info("Batch size {}, state {}", batchInfos.length, batchInfos[0].getState());
```
albertshau, line 200: is there always just one batch? If so this should really be split on the batch results and not on the batch itself. In other words, wait for the results here, get the header, then create a split per result. 

This can be a future improvement, don't need to do it in this PR, which is just fixing bugs.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: fixed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: thanks, I have added a loop + unit test for empty data and result containing only a header.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
rmstar: I'll fix this in a separate PR

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
```diff
@@ -143,50 +170,59 @@ public void close() throws IOException {
...
+      } catch (AsyncApiException e) {
+        if (i == SalesforceSourceConstants.GET_BATCH_RESULTS_TRIES - 1) {
+          throw e;
         }
+        LOG.warn("Failed to get info for batch {}. Will retry after some time.", batchId, e);
+        continue;
+      }
 
-        return new SequenceInputStream(Collections.enumeration(streams));
+      if (info.getState() == BatchStateEnum.Completed) {
```
albertshau, line 224: may be helpful to add a debug message in the else clause which include the batch id and state (and whatever else might be useful), so that the logs have some indication of what the plugin is doing.

albertshau (approved):
lgtm, with one suggestion for additional logging. Please squash and commit.

Also, please be sure to cherry-pick this fix onto develop as well.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBulkRecordReader.java
rmstar: done

PR #90, opened 2021-09-16
Plugin fixes

rmstar:
- Each query result has a header, not just the first one
- Updated the plugin documnetation to indicate that PK chunking should not be used when the query filters out most of the data from a large table.

### DIFF
--- a/docs/Salesforce-batchsource.md
+++ b/docs/Salesforce-batchsource.md
@@ -163,6 +163,9 @@ Primary key (PK) Chunking splits query on large tables into chunks based on the 
 
 Salesforce recommends that you enable PK chunking when querying tables with more than 10 million records or when a bulk query consistently times out. 
 However, the effectiveness of PK chunking depends on the specifics of the query and the queried data.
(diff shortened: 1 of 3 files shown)


docs/Salesforce-batchsource.md
```diff
@@ -163,6 +163,7 @@ Primary key (PK) Chunking splits query on large tables into chunks based on the
 
 Salesforce recommends that you enable PK chunking when querying tables with more than 10 million records or when a bulk query consistently times out. 
 However, the effectiveness of PK chunking depends on the specifics of the query and the queried data.
+We do not recommend enabling PK chunking when querying a large table and filtering out most of the data.
```
albertshau, line 166: Should also mention that each chunk counts as a separate query against the quota, so using PK chunking on a large table can end up counting as thousands of queries against the quota.

albertshau (approved):

docs/Salesforce-batchsource.md
rmstar: thanks, done

albertshau:
please be sure to cherry-pick this onto develop too.

PR #95, opened 2021-10-25
Fix bulk close error "Closing already closed Job not allowed" when pk chunking was enabled

snehalakshmisha:
Resolve jira - https://cdap.atlassian.net/browse/PLUGIN-777 which would cause the Salesforce plugin to fail with "Closing already closed Job not allowed" error

The fix was to create and close the bulk connection only once by moving it out from InputFormatProvider class to prepareRun() and onRunFinish() methods of the SalesforceBatchSource class. This also means we now generate the splits in prepareRun() and send the serialized splits to InputFormat.

Testing:
SalesforceBatchSource, SalesforceBatchMultiSource and SalesforceMultiObjectSource plugins were tested with an sObject table with ~4M records.
In all the three testing cases, the pipeline succeeded and the logs did not show any errors.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchMultiSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchMultiSource.java
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.salesforce.plugin.source.batch;
 
+import com.sforce.async.BulkConnection;
 import com.sforce.ws.ConnectionException;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
@@ -31,8 +32,12 @@ import io.cdap.cdap.etl.api.action.SettableArguments;
 import io.cdap.cdap.etl.api.batch.BatchRuntimeContext;
 import io.cdap.cdap.etl.api.batch.BatchSource;
 import io.cdap.cdap.etl.api.batch.BatchSourceContext;
+import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
+import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
+import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSplitUtil;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -54,6 +59,8 @@ public class SalesforceBatchMultiSource extends BatchSource<Schema, Map<String, 
 
   private final SalesforceMultiSourceConfig config;
   private MapToRecordTransformer transformer;
+  private List<String> jobIds = new ArrayList<>();
+  private AuthenticatorCredentials authenticatorCredentials;
 
   public SalesforceBatchMultiSource(SalesforceMultiSourceConfig config) {
     this.config = config;
@@ -81,8 +88,19 @@ public class SalesforceBatchMultiSource extends BatchSource<Schema, Map<String, 
       (sObjectName, sObjectSchema) -> arguments.set(MULTI_SINK_PREFIX + sObjectName, sObjectSchema.toString()));
 
     String sObjectNameField = config.getSObjectNameField();
+    authenticatorCredentials = SalesforceConnectionUtil.getAuthenticatorCredentials(config.getUsername(),
+                                                                                    config.getPassword(),
+                                                                                    config.getConsumerKey(),
+                                                                                    config.getConsumerSecret(),
+                                                                                    config.getLoginUrl());
+    BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+    List<SalesforceSplit> querySplits = queries.parallelStream()
+      .map(query -> SalesforceSplitUtil.getQuerySplits(query, bulkConnection, false))
+      .flatMap(Collection::stream).collect(Collectors.toList());
+    // store the jobIds so be used in onRunFinish() to close the connections
+    querySplits.parallelStream().forEach(salesforceSplit -> jobIds.add(salesforceSplit.getJobId()));
     context.setInput(Input.of(config.referenceName, new SalesforceInputFormatProvider(
-      config, queries, getSchemaWithNameField(sObjectNameField, schemas), sObjectNameField)));
+      config, getSchemaWithNameField(sObjectNameField, schemas), querySplits, sObjectNameField)));
     /* TODO PLUGIN-510
      *  As part of [CDAP-16290], recordLineage function was introduced with out implementation.
      *  To avoid compilation errors the code block is commented for future fix.
@@ -92,6 +110,12 @@ public class SalesforceBatchMultiSource extends BatchSource<Schema, Map<String, 
                     "Read", "Read from Salesforce MultiObjects.");
     }
      */
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    super.onRunFinish(succeeded, context);
+    SalesforceSplitUtil.closeJobs(jobIds, authenticatorCredentials);
   }
 
   @Override
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
@@ -16,7 +16,9 @@
 package io.cdap.plugin.salesforce.plugin.source.batch;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
+import com.sforce.async.BulkConnection;
 import com.sforce.ws.ConnectionException;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
@@ -33,10 +35,14 @@ import io.cdap.cdap.etl.api.batch.BatchSource;
 import io.cdap.cdap.etl.api.batch.BatchSourceContext;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.salesforce.SObjectDescriptor;
+import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
 import io.cdap.plugin.salesforce.SalesforceSchemaUtil;
+import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSourceConstants;
+import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSplitUtil;
 
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
@@ -49,12 +55,13 @@ import java.util.stream.Collectors;
 @Description("Read data from Salesforce.")
 public class SalesforceBatchSource extends BatchSource<Schema, Map<String, String>, StructuredRecord> {
 
-
   public static final String NAME = "Salesforce";
 
   private final SalesforceSourceConfig config;
   private Schema schema;
   private MapToRecordTransformer transformer;
+  private List<String> jobIds = new ArrayList<>();
+  private AuthenticatorCredentials authenticatorCredentials;
 
   public SalesforceBatchSource(SalesforceSourceConfig config) {
     this.config = config;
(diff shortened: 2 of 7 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -86,25 +110,50 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
 
   @Override
   public void prepareRun(BatchSourceContext context) {
-    FailureCollector collector = context.getFailureCollector();
-    config.validate(collector); // validate when macros are already substituted
-    collector.getOrThrowException();
+    try {
+      FailureCollector collector = context.getFailureCollector();
+      config.validate(collector); // validate when macros are already substituted
+      collector.getOrThrowException();
```
snehalakshmisha: Move it out of try-catch

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -35,17 +35,15 @@
   public static final String PROPERTY_PK_CHUNK_ENABLE_NAME = "enablePKChunk";
   public static final String PROPERTY_CHUNK_SIZE_NAME = "chunkSize";
   public static final String PROPERTY_PARENT_NAME = "parent";
+  public static final String PROPERTY_QUERY_SPLITS = "querySplits";
```
rmstar: Nit: name this similar to other config params passed to the input format. Something like mapred.salesforce.input.query.split

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -35,17 +35,15 @@
...
 
-  public static final String PROPERTY_WHITE_LIST = "whiteList";
-  public static final String PROPERTY_BLACK_LIST = "blackList";
+  public static final String PROPERTY_ALLOW_LIST = "allowList";
```
rmstar: you would also need to update https://github.com/data-integrations/salesforce/blob/develop/widgets/SalesforceMultiObjects-batchsource.json#L64. Suggest doing renaming changes in a separate PR as it's unrelated to this fix.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -100,11 +124,30 @@ public void prepareRun(BatchSourceContext context) {
       Preconditions.checkNotNull(schema.getFields()).stream()
         .map(Schema.Field::getName)
         .collect(Collectors.toList()));
+      String query = config.getQuery(context.getLogicalStartTime());
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+      List<SalesforceSplit> querySplits = generateSalesforceSplits(query);
+      context.setInput(Input.of(config.referenceName, new SalesforceInputFormatProvider(
+      config, Collections.singletonList(query), ImmutableMap.of(
```
rmstar: fix indent

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchMultiSource.java
```diff
@@ -82,7 +82,7 @@ public void prepareRun(BatchSourceContext context) throws ConnectionException {
 
     String sObjectNameField = config.getSObjectNameField();
     context.setInput(Input.of(config.referenceName, new SalesforceInputFormatProvider(
-      config, queries, getSchemaWithNameField(sObjectNameField, schemas), sObjectNameField)));
+      config, queries, getSchemaWithNameField(sObjectNameField, schemas), null, sObjectNameField)));
```
rmstar: Looks like null querySplits is not handled in SalesforceInputFormatProvider

src/main/java/io/cdap/plugin/salesforce/SalesforceConnectionUtil.java
```diff
@@ -42,6 +42,22 @@ public static PartnerConnection getPartnerConnection(AuthenticatorCredentials cr
...
+   * Creates {@link AuthenticatorCredentials} instance based on given parameters.
+   *
+   * @param username Salesforce username
+   * @param password Salesforce password
+   * @param consumerKey Salesforce consumer key
+   * @param consumerSecret Salesforce consumer secret
+   * @param loginUrl Salesforce authentication url
+   * @return authenticator credentials
+   */
+  public static AuthenticatorCredentials getAuthenticatorCredentials(String username, String password,
```
rmstar: Looks like the compile error is fixed in another PR - https://github.com/data-integrations/salesforce/pull/96.
I think it would be easier if we merge that and just fix the bulk close error in this PR - avoid merge conflicts when we cherrypick to release/1.3 branch.

rmstar (commented):
Please make sure the following work correctly:
- PK chunking disabled
- PK chunking enabled
- multi source
Thanks!

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
snehalakshmisha: Thanks. I also see similar terms in the UI as well as in a couple other files. I did not change that because I wasn't sure if it wouldn't break anything. Will drop these changes for now

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -65,30 +48,16 @@
...
-      chunkHeaderValues.add(String.format(SalesforceSourceConstants.HEADER_VALUE_PK_CHUNK, chunkSize));
-      if (!Strings.isNullOrEmpty(parent)) {
-        chunkHeaderValues.add(String.format(SalesforceSourceConstants.HEADER_PK_CHUNK_PARENT, parent));
-      }
-
-      bulkConnection.addHeader(SalesforceSourceConstants.HEADER_ENABLE_PK_CHUNK, String.join(";", chunkHeaderValues));
-    }
-
+    List<SalesforceSplit> querySplits = GSON.fromJson(
+      configuration.get(SalesforceSourceConstants.CONFIG_QUERY_SPLITS), QUERY_SPLITS_TYPE);
```
rmstar, line 55: Need to handle null querySplits. Might be easier if we just passed an empty list

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,209 @@
...
+   * @return an array of batches
+   * @throws AsyncApiException  if there is an issue creating the job
+   * @throws IOException failed to close the query
+   */
+  public static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query, boolean enablePKChunk)
+    throws AsyncApiException, IOException {
+
+    SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
+    JobInfo job = SalesforceBulkUtil.createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query, null);
+    //jobIds.add(job.getId());
```
rmstar: remove commented code

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,209 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Utility class which provides methods to generate Salesforce splits for a query.
+ */
+public class SalesforceSplitUtil {
```
rmstar: final class

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -63,34 +45,15 @@
...
-      bulkConnection.addHeader(SalesforceSourceConstants.HEADER_ENABLE_PK_CHUNK, String.join(";", chunkHeaderValues));
-    }
-
-    return queries.parallelStream()
-      .map(query -> getQuerySplits(query, bulkConnection, enablePKChunk))
-      .flatMap(Collection::stream)
-      .collect(Collectors.toList());
+    List<SalesforceSplit> querySplits = GSON.fromJson(
+      configuration.get(SalesforceSourceConstants.CONFIG_QUERY_SPLITS), QUERY_SPLITS_TYPE);
+    return querySplits.parallelStream().collect(Collectors.toList());
```
rmstar, line 57: why create a stream if you're not doing anything with it?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -40,12 +40,9 @@
   public static final String PROPERTY_BLACK_LIST = "blackList";
   public static final String PROPERTY_SOBJECT_NAME_FIELD = "sObjectNameField";
 
-  public static final String CONFIG_QUERIES = "mapred.salesforce.input.queries";
   public static final String CONFIG_SCHEMAS = "mapred.salesforce.input.schemas";
+  public static final String CONFIG_QUERY_SPLITS = "mapred.salesforce.input.query.split";
```
rmstar: nit: split -> splits

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
snehalakshmisha: Added a comment as discussed offline

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  public static void closeJobs(List<String> jobIds, AuthenticatorCredentials authenticatorCredentials) {
+    jobIds.parallelStream().forEach(jobId -> {
+      BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+      if (bulkConnection != null) {
+        try {
+          SalesforceBulkUtil.closeJob(bulkConnection, jobId);
+        } catch (AsyncApiException e) {
+          throw new RuntimeException(e);
```
rmstar: If there are multiple jobs and closing the first one throws an exception, you'll fail to close the remaining jobs.
To handle that case, throw the exception after closing all jobs. If there are multiple exceptions you can addSuppressed()

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+      }
+    }
+    throw new BulkAPIBatchException("Timeout waiting for batch results", initialBatchInfo);
+  }
+
+  public static void closeJobs(List<String> jobIds, AuthenticatorCredentials authenticatorCredentials) {
+    jobIds.parallelStream().forEach(jobId -> {
+      BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+      if (bulkConnection != null) {
```
rmstar: move this out of the forEach

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,208 @@
...
+   * Start batch job of reading a given guery result.
+   *
+   * @param bulkConnection bulk connection instance
+   * @param query a SOQL query
+   * @param enablePKChunk enable PK Chunk
+   * @return an array of batches
+   * @throws AsyncApiException  if there is an issue creating the job
+   * @throws IOException failed to close the query
+   */
+  public static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query, boolean enablePKChunk)
```
rmstar: Does this need to be public? Is it used outside SalesforceSplitUtil?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -0,0 +1,208 @@
...
+    SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
+    JobInfo job = SalesforceBulkUtil.createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query, null);
+    BatchInfo batchInfo;
+    try (ByteArrayInputStream bout = new ByteArrayInputStream(query.getBytes())) {
+      batchInfo = bulkConnection.createBatchFromStream(job, bout);
+    }
+
+    if (enablePKChunk) {
+      LOG.info("PKChunk is enabled, returning something now");
+      return waitForBatchChunks(bulkConnection, job.getId(), batchInfo.getId());
```
rmstar, line 110: Change "return something now" to something like 'waiting for batch chunks. Also, this and the next log msg can logged at debug log level

snehalakshmisha: Oops, I meant to remove this. Had this only for logging

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -97,14 +104,32 @@ public void prepareRun(BatchSourceContext context) {
     LineageRecorder lineageRecorder = new LineageRecorder(context, config.referenceName);
     lineageRecorder.createExternalDataset(schema);
     lineageRecorder.recordRead("Read", "Read from Salesforce",
-      Preconditions.checkNotNull(schema.getFields()).stream()
-        .map(Schema.Field::getName)
-        .collect(Collectors.toList()));
-
+    Preconditions.checkNotNull(schema.getFields()).stream()
```
rmstar: Looks like incorrect indent, please revert

rmstar (approved):

rmstar (approved):

PR #98, opened 2021-11-29
Bugfix-Plugin-952: Salesforce Streaming Source plugin does not read f…

saiyaswanth-cloudsufi:
…rom topic created by CDF

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
@@ -51,7 +51,7 @@ public class SalesforcePushTopicListener {
    * Timeout of 110 seconds is enforced by Salesforce Streaming API and is not configurable.
    * So we enforce the same on client.
    */
-  private static final int CONNECTION_TIMEOUT_SECONDS = 110;
+  private static final int CONNECTION_TIMEOUT_SECONDS = 1100;
   private static final long HANDSHAKE_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(110);
 
   private static final int HANDSHAKE_CHECK_INTERVAL_MS = 1000;


rmstar (commented):
Can you update the bug description to provide some more context. Why are we bumping up the connection timeout and how does that fix the issue?

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -51,7 +51,7 @@
    * Timeout of 110 seconds is enforced by Salesforce Streaming API and is not configurable.
```
rmstar, line 51: Comment indicates that timeout is not configurable.. how does bumping it up to 1100 seconds help here?

saiyaswanth-cloudsufi: @rmstar HttpClient is expecting time is in milli seconds, but the connection timeout which is mentioned is very less and unable to connect. Due to this it is causing timeout exception.

rmstar: So there are 2 timeouts - connection timeout and handshake timeout. Can you update the comment to indicate why you're setting the connection timeout to 1100 seconds, but handshake timeout is 110 seconds (i.e why 1 timeout is 10x the other timeout)?

saiyaswanth-cloudsufi: @rmstar HANDSHAKE_TIMEOUT_MS and HANDSHAKE_CHECK_INTERVAL_MS is to wait for sometime to wait after connecting giving a request to the salesforce basically it is waiting whether the handshake happened. But the CONNECTION_TIMEOUT_SECONDS is to give a request call to the salesforce using the HttpClient, as it is having less time, it is unable to complete the request by that time.

vikasrathee-cs: separate PR has been raised for this by Sanchit. Please discard this PR

PR #115, opened 2022-01-27
add queryAll SOQL operation

jster1357:
add queryAll SOQL operation

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.salesforce.plugin.source.batch;
 
+import com.sforce.async.OperationEnum;
 import com.sforce.ws.ConnectionException;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
@@ -75,6 +76,14 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
   @Macro
   private String offset;
 
+  @Name(SalesforceSourceConstants.PROPERTY_OPERATION)
+  @Description("If set to query, the query result will only return current rows. If set to queryAll, " +
+          "all records, including deletes will be sourced")
+  @Nullable
+  private String operation;
+
+  private static final String DEFAULT_OPERATION = "query";
+
   protected SalesforceBaseSourceConfig(String referenceName,
                                        String consumerKey,
                                        String consumerSecret,
@@ -85,12 +94,14 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
                                        @Nullable String datetimeBefore,
                                        @Nullable String duration,
                                        @Nullable String offset,
-                                       @Nullable String securityToken) {
+                                       @Nullable String securityToken,
+                                       @Nullable String operation) {
     super(referenceName, consumerKey, consumerSecret, username, password, loginUrl, securityToken);
     this.datetimeAfter = datetimeAfter;
     this.datetimeBefore = datetimeBefore;
     this.duration = duration;
     this.offset = offset;
+    this.operation = operation;
   }
 
   public Map<ChronoUnit, Integer> getDuration() {
@@ -129,6 +140,11 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
     }
     try {
       validateRangeFilterProperty(SalesforceSourceConstants.PROPERTY_OFFSET, getOffset());
+    } catch (InvalidConfigException e) {
+      collector.addFailure(e.getMessage(), null).withConfigProperty(e.getProperty());
+    }
+    try {
+      validateOperationProperty(SalesforceSourceConstants.PROPERTY_OPERATION, getOperation());
     } catch (InvalidConfigException e) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(e.getProperty());
     }
@@ -189,7 +205,6 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
     return filterDescriptor;
   }
 
-  @Nullable
   private void validateIntervalFilterProperty(String propertyName, String datetime) {
     if (containsMacro(propertyName)) {
       return;
@@ -200,6 +215,16 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
       throw new InvalidConfigException(
         String.format("Invalid SObject '%s' value: '%s'. Value must be in Salesforce Date Formats. For example, "
                         + "2019-01-01T23:01:01Z", propertyName, datetime), propertyName);
+    }
+  }
+
+  private void validateOperationProperty(String propertyName, String operation) {
+    try {
+      OperationEnum.valueOf(operation);
+    } catch (InvalidConfigException e) {
+      throw new InvalidConfigException(
+              String.format("Invalid Query Operation: '%s' value: '%s'. Valid values are query and queryAll.",
+                      propertyName, operation), propertyName);
     }
   }
 
@@ -268,5 +293,9 @@ public abstract class SalesforceBaseSourceConfig extends BaseSalesforceConfig {
   @Nullable
   private ZonedDateTime parseDatetime(String datetime) throws DateTimeParseException {
     return StringUtils.isBlank(datetime) ? null : ZonedDateTime.parse(datetime, DateTimeFormatter.ISO_DATE_TIME);
+  }
+
+  public String getOperation() {
+    return operation == null ? DEFAULT_OPERATION : operation;
   }
 }
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceMultiSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceMultiSourceConfig.java
@@ -81,9 +81,10 @@ public class SalesforceMultiSourceConfig extends SalesforceBaseSourceConfig {
                               @Nullable String whiteList,
                               @Nullable String blackList,
                               @Nullable String sObjectNameField,
-                              @Nullable String securityToken) {
+                              @Nullable String securityToken,
+                              @Nullable String operation) {
     super(referenceName, consumerKey, consumerSecret, username, password, loginUrl,
-          datetimeAfter, datetimeBefore, duration, offset, securityToken);
+          datetimeAfter, datetimeBefore, duration, offset, securityToken, operation);
     this.whiteList = whiteList;
     this.blackList = blackList;
     this.sObjectNameField = sObjectNameField;
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
@@ -23,6 +23,7 @@ import com.sforce.async.JobInfo;
 import com.sforce.async.JobStateEnum;
 import com.sforce.async.OperationEnum;
 import io.cdap.plugin.salesforce.BulkAPIBatchException;
+import io.cdap.plugin.salesforce.InvalidConfigException;
 import io.cdap.plugin.salesforce.SObjectDescriptor;
 import io.cdap.plugin.salesforce.SalesforceBulkUtil;
 import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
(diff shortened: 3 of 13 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -75,6 +76,11 @@
   @Macro
   private String offset;
 
+  @Name(SalesforceSourceConstants.PROPERTY_OPERATION)
+  @Description("If set to query, the query result will only return current rows. If set to queryAll, " +
+          "all records, including deletes will be sourced")
+  private String operation;
```
rmstar, line 83: should this param be macro enabled?
Also, seems like we should support null operation for backward compatibility, so let's make it nullable and default to 'query' when operation is not specified

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -269,4 +276,17 @@ private int parseUnitValue(String propertyName, String value) {
   private ZonedDateTime parseDatetime(String datetime) throws DateTimeParseException {
     return StringUtils.isBlank(datetime) ? null : ZonedDateTime.parse(datetime, DateTimeFormatter.ISO_DATE_TIME);
   }
+
+  public String getOperation() {
+    return operation;
+  }
+
+  public OperationEnum getOperationEnum() {
```
rmstar: where is this called?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -96,11 +97,13 @@
...
    */
-  private static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query, boolean enablePKChunk)
+  private static BatchInfo[] runBulkQuery(BulkConnection bulkConnection, String query,
+                                          boolean enablePKChunk, String operation)
     throws AsyncApiException, IOException {
 
     SObjectDescriptor sObjectDescriptor = SObjectDescriptor.fromQuery(query);
-    JobInfo job = SalesforceBulkUtil.createJob(bulkConnection, sObjectDescriptor.getName(), OperationEnum.query, null);
+    JobInfo job = SalesforceBulkUtil.createJob(bulkConnection, sObjectDescriptor.getName(),
+            OperationEnum.valueOf(operation), null);
```
rmstar: handle null operation, annotate with @Nullable

jster1357:
JIRA for the change: https://cdap.atlassian.net/browse/PLUGIN-701

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: I have this set as a radio button, defaulted to "query"...does this work?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: older code...I'll remove

src/test/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfigBuilder.java
```diff
@@ -123,9 +124,14 @@ public SalesforceSourceConfigBuilder setParent(String parent) {
     return this;
   }
 
+  public SalesforceSourceConfigBuilder setOperation(String operation) {
```
snehalakshmisha, line 127: Where is this method called?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
rmstar: That works when the user is building a pipeline in pipeline studio. 
But we also need to support the case where a pipeline JSON with an older version of salesforce plugin is imported. 
The old pipeline json won't have 'operation' config.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
snehalakshmisha: You've defined SalesforceBaseSourceConfig.getOperationEnum() which implies `Operation.valueOf(enum)` can throw exceptions and that should probably be handled here as well

src/test/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfigBuilder.java
jster1357: I haven't built any test cases for it. It was erroring out when I tried to build it so I added this get past the build.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -269,4 +277,10 @@ private int parseUnitValue(String propertyName, String value) {
   private ZonedDateTime parseDatetime(String datetime) throws DateTimeParseException {
     return StringUtils.isBlank(datetime) ? null : ZonedDateTime.parse(datetime, DateTimeFormatter.ISO_DATE_TIME);
   }
+
+  public String getOperation() {
+    return operation == null ? "query" : operation;
```
snehalakshmisha: `query` should be a constant string

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -269,4 +277,10 @@ private int parseUnitValue(String propertyName, String value) {
   private ZonedDateTime parseDatetime(String datetime) throws DateTimeParseException {
     return StringUtils.isBlank(datetime) ? null : ZonedDateTime.parse(datetime, DateTimeFormatter.ISO_DATE_TIME);
   }
+
+  public String getOperation() {
```
rmstar, line 298: add code in validate() to check the operation is either query or queryAll. Add failure to failure collector if the operation is invalid

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: added a constant string

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
jster1357: added new method that deals with this

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: got it. I've made it nullable

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: so add similar logic to whats being done with validateFilters()?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.salesforce.plugin.source.batch;
 
+import com.sforce.async.OperationEnum;
```
snehalakshmisha, line 18: If this isn't used in the class anymore it should be remvoed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -75,6 +76,14 @@
   @Macro
   private String offset;
 
+  @Name(SalesforceSourceConstants.PROPERTY_OPERATION)
+  @Description("If set to query, the query result will only return current rows. If set to queryAll, " +
```
snehalakshmisha, line 80: nit: Is there a way to highlight `query` and `queryAll` to indicate that it is a keyword?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -75,6 +76,14 @@
...
+          "all records, including deletes will be sourced")
+  @Nullable
+  private String operation;
+
+  private static final String default_operation = "query";
```
snehalakshmisha: `default_operation` should be in upper case

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceMultiSourceConfig.java
```diff
@@ -81,7 +81,8 @@ public SalesforceMultiSourceConfig(String referenceName,
                               @Nullable String whiteList,
                               @Nullable String blackList,
                               @Nullable String sObjectNameField,
-                              @Nullable String securityToken) {
+                              @Nullable String securityToken,
+                              @Nullable String operation) {
```
snehalakshmisha, line 85: If we are passing operation in the constructor, we must want to use that value somewhere? Either set the value of the field here or pass it to super().

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -22,11 +22,7 @@
 import com.sforce.async.JobInfo;
 import com.sforce.async.JobStateEnum;
 import com.sforce.async.OperationEnum;
-import io.cdap.plugin.salesforce.BulkAPIBatchException;
-import io.cdap.plugin.salesforce.SObjectDescriptor;
-import io.cdap.plugin.salesforce.SalesforceBulkUtil;
-import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
-import io.cdap.plugin.salesforce.SalesforceQueryUtil;
+import io.cdap.plugin.salesforce.*;
```
snehalakshmisha: We should not have * imports. Import only the classes you need.
You can do a `mvn clean package` after you address any comments for sanity check, that will catch issues some issues like this for you

jster1357: Thanks. I fixed this on my latest build.

snehalakshmisha: You can turn off automatic wildcard imports - https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports-

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: done

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
jster1357: done

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceMultiSourceConfig.java
jster1357: added it to super

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
rmstar: yes. convert operation string to enum. If that fails (exception is thrown), then add a failure to the failure colllector.

jster1357: I just pushed some code that does validation.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -203,6 +218,16 @@ private void validateIntervalFilterProperty(String propertyName, String datetime
     }
   }
 
+  private void validateOperationProperty(String propertyName, String operation) {
+    try {
+      OperationEnum.valueOf(operation);
+    } catch (InvalidConfigException e) {
+      throw new InvalidConfigException(
+              String.format("Invalid Query Operation: '%s' value: '%s'. Valid values are query and queryAll.",
```
rmstar, line 226: change the string to -
String.format("Invalid Query Operation: '%s'. Valid values are query and queryAll.", operation)

rmstar (approved):
please squash commits after addressing the minor comment

PR #117, opened 2022-02-02
Bugfix plugin 1033

kunalsharma-repo:


rmstar (commented):
Also add 'Scope' keyword

kunalsharma-repo:
> Also add 'Scope' keyword

Added Scope keyword

src/test/java/io/cdap/plugin/salesforce/parser/SalesforceQueryParserTest.java
```diff
@@ -266,4 +266,10 @@ public void testSubQuery() {
       Collections.singletonList(new SObjectDescriptor("Contacts", fields)),
       result.getChildSObjects());
   }
+
+  @Test
+  public void testParseFailure() {
```
rmstar: Rename to something like testKeywordInQuery

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -158,6 +158,30 @@ public SObjectDescriptor visitStatement(SOQLParser.StatementContext ctx) {
...
+        throw new SOQLParsingException("Invalid column name: " + ctx.getText());
+      }
+
+      if (nameParts.size() > 1) {
+        // if field contains alias, remove it to get actual field name
+        // Example:
+        // select Opportunity.Name from Opportunity -> Name
+        // select o.Name from Opportunity o -> Name
+        String topParent = nameParts.get(0);
+        if (topParent.equalsIgnoreCase(objectName) || topParent.equals(objectAlias)) {
```
rmstar, line 177: why do we ignore case when checking equality with objectName, but not with objectAlias?

src/test/java/io/cdap/plugin/salesforce/parser/SalesforceQueryParserTest.java
kunalsharma-repo: Method renamed from testParseFailure to testKeywordInQuery

src/main/antlr4/soql/SOQL.g4
```diff
@@ -61,6 +61,10 @@ keyword             : GROUP
                     | CATEGORY
                     ;
 
+selectKey           : DATA
+                    | SCOPE
```
snehalakshmisha, line 65: We are only fixing the issue for Data and Scope keywords. Do we also want to consider other keywords like end, etc?

vikasrathee-cs:
Closing this PR as it is having cla issue.
Raised a new PR.
https://github.com/data-integrations/salesforce/pull/131

PR #118, opened 2022-02-04
Fixed instances of getting credentials through a helper that assumed u-name/password

ssmith-google:
This helper method assumed credentials were of the form username/password/etc.
This failed for pipelines utilizing the OAuth flow support.

Removed the aforementioned helper method, as all instances of it being
used could instead get the credentials through the base config class.
This config class correctly handles the OAuth flow.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
@@ -26,7 +26,6 @@ import io.cdap.plugin.salesforce.BulkAPIBatchException;
 import io.cdap.plugin.salesforce.InvalidConfigException;
 import io.cdap.plugin.salesforce.SObjectDescriptor;
 import io.cdap.plugin.salesforce.SalesforceBulkUtil;
-import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
 import io.cdap.plugin.salesforce.SalesforceQueryUtil;
 import io.cdap.plugin.salesforce.authenticator.Authenticator;
 import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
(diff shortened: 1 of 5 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -124,22 +123,6 @@
   }
 
 
-  /**
-   * Initializes bulk connection based on given Hadoop credentials configuration.
-   *
-   * @return bulk connection instance
-   */
-  public static BulkConnection getBulkConnection(String username, String password,
```
snehalakshmisha, line 132: Why is this method removed?

snehalakshmisha:
Can you add some details on how you tested the changes?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
ssmith-google: This variant of getBulkConnection appears to be unused and runs into the same issue as the getAuthenticatorCredentials variant I removed: it assumes username/password/etc. and will break if OAuth is used. The other variant takes an AuthenticatorCredentials directly, so will work in either case.

ssmith-google:
> Can you add some details on how you tested the changes?

I created a pipeline with a Salesforce data source using the OAuth flow and verified that it could successfully read in data. I didn't see any existing tests that could be extended to cover these changes, so I welcome any suggestions that you might have. I'm not very familiar with the best practices CDAP uses for testing changes.

snehalakshmisha:
> > Can you add some details on how you tested the changes?
> 
> I created a pipeline with a Salesforce data source using the OAuth flow and verified that it could successfully read in data. I didn't see any existing tests that could be extended to cover these changes, so I welcome any suggestions that you might have. I'm not very familiar with the best practices CDAP uses for testing changes.

Please test both SalesforceBatchSource and SalesforceBatchMultiSource - one pipeline for each. Before and after the code changes the pipeline behaviour should be the same - you should be able to see the same number of records processed, the pipeline should succeed and there shouldn't be any new exceptions or warnings(even if the pipeline succeeds) in the logs.

ssmith-google:
> > > Can you add some details on how you tested the changes?
> > 
> > 
> > I created a pipeline with a Salesforce data source using the OAuth flow and verified that it could successfully read in data. I didn't see any existing tests that could be extended to cover these changes, so I welcome any suggestions that you might have. I'm not very familiar with the best practices CDAP uses for testing changes.
> 
> Please test both SalesforceBatchSource and SalesforceBatchMultiSource - one pipeline for each. Before and after the code changes the pipeline behaviour should be the same - you should be able to see the same number of records processed, the pipeline should succeed and there shouldn't be any new exceptions or warnings(even if the pipeline succeeds) in the logs.

I've created test pipelines for both SalesforceBatchSource and SalesforceBatchMultiSource. Both succeeded with the same number of records processed, and there doesn't appear to be any new messages in the logs.

rmstar (approved):

PR #131, opened 2022-08-01
PLUGIN-1033 : Add Keywords Support for SOQL Query

sgarg-CS:
[`PLUGIN-1033`](https://cdap.atlassian.net/browse/PLUGIN-1033)

### DIFF
--- a/src/main/antlr4/soql/SOQL.g4
+++ b/src/main/antlr4/soql/SOQL.g4
@@ -61,6 +61,12 @@ keyword             : GROUP
                     | CATEGORY
                     ;
 
+selectKey           : DATA
+                    | SCOPE
+                    | END
+                    | OFFSET
+                    ;
+
 SELECT              : S E L E C T ;
 TYPEOF              : T Y P E O F ;
 FROM                : F R O M ;
@@ -235,6 +241,7 @@ fieldList           : '*' # starElement
                     ;
 
 fieldElement        : subquery
+                    | specialFieldName
                     | fieldName
                     | functionCall
                     | typeOfClause
@@ -248,6 +255,10 @@ subquery            : '('
                         (LIMIT numberOfRowsToReturn)?
                         (OFFSET numberOfRowsToSkip)?
                       ')' ;
+
+selectKeyField      : selectKey ('.' selectKey)* ;
+
+specialFieldName    : selectKeyField alias?;
 
 fieldName           : field alias?; // only for aggregate queries
 
--- a/src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
+++ b/src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
@@ -148,7 +148,7 @@ public class SalesforceQueryVisitor extends SOQLBaseVisitor<SObjectDescriptor> {
         // select Opportunity.Name from Opportunity -> Name
         // select o.Name from Opportunity o -> Name
         String topParent = nameParts.get(0);
-        if (topParent.equalsIgnoreCase(objectName) || topParent.equals(objectAlias)) {
+        if (topParent.equals(objectName) || topParent.equals(objectAlias)) {
           // remove alias
           nameParts = nameParts.subList(1, nameParts.size());
         }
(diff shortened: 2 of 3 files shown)


src/test/java/io/cdap/plugin/salesforce/parser/SalesforceQueryParserTest.java
```diff
@@ -266,4 +266,16 @@ public void testSubQuery() {
...
+  public void testKeywordInQuery() {
+    String query = "SELECT Data FROM BackgroundOperationResult";
+    SalesforceQueryParser.getObjectDescriptorFromQuery(query);
+  }
+
+  @Test
+  public void testEndKeywordInQuery() {
+    String query = "SELECT END FROM BackgroundOperationResult";
+    SalesforceQueryParser.getObjectDescriptorFromQuery(query);
+  }
```
rmstar, line 280: Can we also add a test for SCOPE keyword

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -158,6 +158,30 @@ public SObjectDescriptor.FieldDescriptor visitFieldName(SOQLParser.FieldNameCont
...
+        .collect(Collectors.toList());
+
+      if (nameParts.isEmpty()) {
+        throw new SOQLParsingException("Invalid column name: " + ctx.getText());
+      }
+
+      if (nameParts.size() > 1) {
+        // if field contains alias, remove it to get actual field name
+        // Example:
+        // select Opportunity.Name from Opportunity -> Name
```
rmstar, line 174: Is the query case sensitive?
e.g: is `select opportunity.name from opportunity` (all lowercase) a valid query?

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
```diff
@@ -148,7 +148,7 @@ public SObjectDescriptor.FieldDescriptor visitFieldName(SOQLParser.FieldNameCont
         // select Opportunity.Name from Opportunity -> Name
         // select o.Name from Opportunity o -> Name
         String topParent = nameParts.get(0);
-        if (topParent.equalsIgnoreCase(objectName) || topParent.equals(objectAlias)) {
+        if (topParent.equals(objectName) || topParent.equals(objectAlias)) {
```
rmstar, line 151: why do we need this change to it case sensitive?

src/main/antlr4/soql/SOQL.g4
```diff
@@ -61,6 +61,11 @@ keyword             : GROUP
                     | CATEGORY
                     ;
 
+selectKey           : DATA
+                    | SCOPE
+                    | END
```
snehalakshmisha, line 66: On what basis are we selecting the keywords list here?

src/test/java/io/cdap/plugin/salesforce/parser/SalesforceQueryParserTest.java
sgarg-CS: Added. 1ff5de85beff263ec032bc174f18f5fb929cf292

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
sgarg-CS: No, the query is not case-sensitive.

However the SOQL query: `select  opportunity.name from opportunity` will be transformed by the code to `select name from opportunity` to make it valid.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
sgarg-CS: There was a bug in the Salesforce connector:  [PLUGIN-1184](https://cdap.atlassian.net/browse/PLUGIN-1184) in April this year where we added support for the case in-sensitivity in the SOQL query.

Now, it would work both for the lower case as well as upper case characters.

src/main/antlr4/soql/SOQL.g4
sgarg-CS: These are the keywords that we have encountered from customers' usage. There is no specific list, we can add more if you can suggest.

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
rmstar: I see. Can we add a unit test for a case insensitive query with alias like `select o.name from opportunity o`?

src/main/antlr4/soql/SOQL.g4
rmstar: hm it seems like this is not a comprehensive list; we may hit similar issues in the future for other keywords.
Can we accept any identifier or keyword in the select clause?

src/main/java/io/cdap/plugin/salesforce/parser/SalesforceQueryVisitor.java
sgarg-CS: added cbbb7f710fd0d20e036245323bf178ebfc4d5279

src/main/antlr4/soql/SOQL.g4
sgarg-CS: No, only the ones defined in the `SOQL.g4` file.

We went through all the standard objects [here](https://developer.salesforce.com/docs/atlas.en-us.234.0.object_reference.meta/object_reference/sforce_api_objects_list.htm) and found 1 new keyword as a field name that wasn't covered in the code.

Offset in [CommSubscriptionTiming](https://developer.salesforce.com/docs/atlas.en-us.234.0.object_reference.meta/object_reference/sforce_api_objects_commsubscriptiontiming.htm)

Added to the selectKey c6ac3065a0082add66d16d798a24ecd38d2272a7

rmstar (approved):
please squash commits

sgarg-CS:
> please squash commits

Commits have been squashed. Thank you!

PR #134, opened 2022-08-22
[PLUGIN-1184] Salesforce batch source plugin not able to retrieve data when the query is case insensitive

sgarg-CS:
[PLUGIN-1184](https://cdap.atlassian.net/browse/PLUGIN-1184)
Salesforce batch source plugin is not able to retrieve data when the query is case insensitive

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <name>Salesforce plugins</name>
   <groupId>io.cdap.plugin</groupId>
   <artifactId>salesforce-plugins</artifactId>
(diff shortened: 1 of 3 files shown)


rmstar (commented):
Please fix on develop first, then cherrypick to release branch. Version bump will be a separate commit after your cherrypick commit.

sgarg-CS:
Shared new PR after cherry-picking the changes. 
https://github.com/data-integrations/salesforce/pull/135

PR #138, opened 2022-09-27
CDAP-19638: add fqn support for batch plugins

mrahanjam:
This PR adds FQN to the Salesforce batch plugin code.

Jira: [CDAP-19638](https://cdap.atlassian.net/browse/CDAP-19638)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,13 +58,13 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.4.0</cdap.version>
+    <cdap.version>6.8.0-SNAPSHOT</cdap.version>
     <junit.version>4.12</junit.version>
     <commons.io.version>2.6</commons.io.version>
     <commons.lang.version>2.6</commons.lang.version>
     <hadoop.version>2.3.0</hadoop.version>
     <spark2.version>2.1.3</spark2.version>
-    <hydrator.version>2.4.0</hydrator.version>
+    <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
     <commons.version>3.8.1</commons.version>
     <salesforce.api.version>53.0.0</salesforce.api.version>
     <cometd.java.client.version>4.0.0</cometd.java.client.version>
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
@@ -15,11 +15,14 @@
  */
 package io.cdap.plugin.salesforce.plugin;
 
+import com.google.common.base.Strings;
+import com.sforce.soap.partner.PartnerConnection;
 import com.sforce.ws.ConnectionException;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
 import io.cdap.plugin.salesforce.SalesforceConstants;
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -136,7 +136,7 @@
     <dependency>
       <groupId>io.cdap.plugin</groupId>
       <artifactId>hydrator-common</artifactId>
-      <version>${hydrator.version}</version>
+      <version>${cdap.plugin.version}</version>
```
fernst: Why is this change needed? In general, we use `hydrator.version` to refer to the version of the parent hydrator plugin artifact

src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
```diff
@@ -77,6 +80,9 @@ public class BaseSalesforceConfig extends ReferencePluginConfig {
   @Nullable
   private String loginUrl;
 
+  private final String fqnPrefix = "salesforce://";
```
fernst: final variable names should be `UPPERCASE_SEPARATED_BY_UNDERSCORES`

src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
```diff
@@ -125,6 +131,28 @@ public String getLoginUrl() {
...
+  }
+
+  /**
+   * Get fully-qualified name (FQN) for a Salesforce object (FQN format: salesforce://prod/orgId.mySobject).
+   *
+   * @return String fqn
+   */
+  public String getFQN(String orgId, String sObject) {
+    String firstFQNPart = loginUrl.equals(defaultLoginUrl) ? "prod" : "sandbox";
+    return fqnPrefix + firstFQNPart + "/" + orgId + "." + sObject;
```
fernst: Nitpick: use String.format to make the format clearer to understand.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchMultiSource.java
```diff
@@ -103,17 +106,25 @@ public void prepareRun(BatchSourceContext context) throws ConnectionException {
...
-      recordLineage(context, config.referenceName, schema,
-                    "Read", "Read from Salesforce MultiObjects.");
+    String orgId = "unknown";
+    try {
+      orgId = config.getOrgId();
+    } catch (ConnectionException exception) {
+      collector.addFailure("Unable to get organization Id.", "Ensure Credentials are correct.");
     }
-     */
+    Asset asset = Asset.builder(config.getReferenceNameOrNoarmalizedFQN(orgId, sObjectNameField))
```
fernst: Typo: `Noarmalized` should be `Normalized`

fernst (changes requested):
Added some comments/questions

src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
mrahanjam: I had it like that at first, but checkstyle failed when building the module with `Name 'FQN_PREFIX' must match pattern '^[a-z][a-zA-Z0-9]*$'`. I ended up removing the constant as it is not really needed when I used String.format.

fernst: Oh, you need to make this a static variable. Forgot to mention that.

fernst: ` private static final String FQN_PREFIX = "salesforce://";`

src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
```diff
@@ -77,6 +80,8 @@ public class BaseSalesforceConfig extends ReferencePluginConfig {
   @Nullable
   private String loginUrl;
 
+  private final String defaultLoginUrl = "https://login.salesforce.com/services/oauth2/token";
```
fernst: make this `private static final`

mrahanjam: Done

fernst (approved):

PR #139, opened 2022-10-04
verify lower bound of cdap version

mrahanjam:
[PLUGIN-1376](https://cdap.atlassian.net/browse/PLUGIN-1376)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -489,9 +489,9 @@
               <!--
                 Plugins use the new macro substitution that supports object, which is introduced in 6.4.0.
                 The cdap API version that this project depends on is still 6.2.0 since there is no new API needed.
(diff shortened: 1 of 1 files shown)


snehalakshmisha (approved):

pom.xml
```diff
@@ -486,12 +486,8 @@
           <version>1.1.0</version>
           <configuration>
             <cdapArtifacts>
-              <!--
-                Plugins use the new macro substitution that supports object, which is introduced in 6.4.0.
-                The cdap API version that this project depends on is still 6.2.0 since there is no new API needed.
-               -->
```
snehalakshmisha: we probably need to check what this means

PR #143, opened 2022-11-05
PLUGIN-1435 : Salesforce Streaming Bug Fix

sgarg-CS:
[PLUGIN-1435](https://cdap.atlassian.net/browse/PLUGIN-1435)

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
@@ -22,6 +22,8 @@ import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
 import io.cdap.plugin.salesforce.plugin.OAuthInfo;
 import org.awaitility.Awaitility;
 import org.awaitility.core.ConditionTimeoutException;
+import org.cometd.bayeux.Channel;
+import org.cometd.bayeux.client.ClientSessionChannel;
 import org.cometd.client.BayeuxClient;
 import org.cometd.client.transport.ClientTransport;
 import org.cometd.client.transport.LongPollingTransport;
@@ -61,6 +63,7 @@ public class SalesforcePushTopicListener {
 
   private final AuthenticatorCredentials credentials;
   private final String topic;
+  private BayeuxClient bayeuxClient;
 
   private JSONContext.Client jsonContext;
 
@@ -75,13 +78,9 @@ public class SalesforcePushTopicListener {
    */
   public void start() {
     try {
-      BayeuxClient bayeuxClient = getClient(credentials);
-      waitForHandshake(bayeuxClient, HANDSHAKE_TIMEOUT_MS, HANDSHAKE_CHECK_INTERVAL_MS);
-      LOG.debug("Client handshake done");
-      bayeuxClient.getChannel("/topic/" + topic).subscribe((channel, message) -> {
-        messagesQueue.add(jsonContext.getGenerator().generate(message.getDataAsMap()));
-      });
-
+      createSalesforceListener();
+      waitForHandshake();
+      subscribe();
     } catch (Exception e) {
       throw new RuntimeException("Could not start client", e);
     }
@@ -92,7 +91,7 @@ public class SalesforcePushTopicListener {
    * specified wait time if necessary for an element to become available.
    *
    * @param timeout how long to wait before giving up
-   * @param unit timeunit of timeout
+   * @param unit    timeunit of timeout
    * @return the message, or {@code null} if the specified
    * waiting time elapses before an element is available
    * @throws InterruptedException blocking call is interrupted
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +129,95 @@ protected void customize(Request exchange) {
...
     return client;
   }
 
-  private void waitForHandshake(BayeuxClient client,
-                                long timeoutInMilliseconds, long intervalInMilliseconds) {
+  public void createSalesforceListener() {
+    try {
+      bayeuxClient = getClient(credentials);
+    } catch (Exception exception) {
+      LOG.error("Exception while creating bayeux client ", exception);
```
rmstar: if you catch exception here you'll get a nullpointerexception on the next line. Instead, don't catch exceptions here, let the caller handle it

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +129,95 @@ protected void customize(Request exchange) {
...
+      ((ClientSessionChannel.MessageListener) (channel, message) -> {
+
+        boolean success = message.isSuccessful();
+        if (!success) {
+          String error = (String) message.get("error");
+          if (error != null) {
+            LOG.error("Error during HANDSHAKE: {}", error);
+          }
+
+          Exception exception = (Exception) message.get("exception");
```
rmstar: How do we know that message contains `error`, `exception` keys? Is it documented somewhere?
Also, it would be safer to check if  message.get("exception") instanceof Exception before casting

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +129,95 @@ protected void customize(Request exchange) {
...
+          if (exception != null) {
+            LOG.error("Exception during HANDSHAKE", exception);
```
rmstar: we should reconnect on error. Otherwise we'll hit the same issue if there are transient errors

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +129,95 @@ protected void customize(Request exchange) {
...
+          Map<String, Object> advice = message.getAdvice();
+
+          if (error != null) {
+            LOG.error("Error during CONNECT: {}", error);
+            LOG.debug("Advice during CONNECT: {}", advice);
+          }
+          if (advice.get("reconnect").equals("handshake")) {
+            LOG.debug("Reconnecting to Salesforce Push Topic");
+            reconnectToTopic();
+          }
```
rmstar, line 180: why aren't we reconnecting when there are other types of errors?

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +129,95 @@ protected void customize(Request exchange) {
...
+        }
+      });
+
+  }
+
+  public void reconnectToTopic() {
+    disconnectStream();
+    createSalesforceListener();
+    waitForHandshake();
+    LOG.debug("start() Client handshake done");
```
rmstar: nit: remove start().
You can move this log line to the end of waitForHandshake(), which avoids duplicating it here and in start()

rmstar (commented):
Please look into adding a unit test.
For example, refer to Pub/Sub tests - https://github.com/data-integrations/google-cloud/blob/develop/src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiverMessagePullingTest.java, https://github.com/data-integrations/google-cloud/blob/develop/src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiverSubscriptionTest.java

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: Done. https://github.com/data-integrations/salesforce/commit/547e2e9ff762dc326333289005a82779f0cce293

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: We specifically receive the following error message after the expiration of the 'sfdc-stream' cookie managed by the Salesforce load balancer; it has a 3hr expiration time, with the advice to reconnect using a handshake
[Mulesoft Help Center](https://help.mulesoft.com/s/article/Couldn-t-connect-to-Salesforce-streaming-server-because-a-403-error-has-been-returned-error-403-Unknown-client-Warning-messages-in-log)
```
{
  "error": "403::Unknown client",
  "successful": false,
  "advice": {
    "interval": 0,
    "reconnect": "handshake"
  }
}
```

The other types of error messages do not contain this reconnect advice.
A list of errors can be found on the Page No. 48 in this [document](https://blog.bessereau.eu/assets/pdfs/api_streaming.pdf)

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: We're reconnecting only on the error(s) which have `handshake` as the value in the `reconnect` key in the `advice` object.

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: https://developer.salesforce.com/docs/atlas.en-us.api_streaming.meta/api_streaming/streaming_error_codes.htm
Added the same to the code. https://github.com/data-integrations/salesforce/pull/143/commits/5d519e00f9b98a841f488106261fdebeda9b9001

Also added the check if the `message.get("exception")` is an instance of the `Exception` class https://github.com/data-integrations/salesforce/pull/143/commits/547e2e9ff762dc326333289005a82779f0cce293

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: Removed the catch block. Added exception to the method signature. 547e2e9ff762dc326333289005a82779f0cce293

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +128,103 @@ protected void customize(Request exchange) {
...
+            LOG.error("Error during HANDSHAKE: {}", error);
+            throw new RuntimeException("Error in meta handshake , errorMessage:" + error);
+          }
+          if (message.get("exception") instanceof Exception) {
+            Exception exception = (Exception) message.get("exception");
+            if (exception != null) {
+              LOG.error("Exception during HANDSHAKE", exception);
+              throw new RuntimeException("Error in meta handshake ", exception);
+            }
+          }
```
rmstar, line 150: Is it possible that `!message.isSuccessful()` and message doesn't contain `error` or `exception`?

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
```diff
@@ -129,20 +128,103 @@ protected void customize(Request exchange) {
...
+    bayeuxClient.getChannel(Channel.META_SUBSCRIBE).addListener(
+      (ClientSessionChannel.MessageListener) (channel, message) -> {
+
+        boolean success = message.isSuccessful();
+        if (!success) {
+          String error = (String) message.get("error");
+          if (error != null) {
+            LOG.error("Error during SUBSCRIBE: " + error);
+            throw new RuntimeException("Error in meta subscribe , errorMessage:" + error);
+          }
```
rmstar, line 194: what happens if the success = false and message doesn't contain `error`?

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: That is unlikely to happen. We haven't come across any message that is not successful and does not contain any `error` or `exception`

All messages having `"successful": false` contain the `error` key.

Some common Streaming API Error Messages: https://developer.salesforce.com/docs/atlas.en-us.api_streaming.meta/api_streaming/streaming_handling_errors.htm

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforcePushTopicListener.java
sgarg-CS: The pipeline will continue to run as usual.

sgarg-CS:
> Please look into adding a unit test. For example, refer to Pub/Sub tests - https://github.com/data-integrations/google-cloud/blob/develop/src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiverMessagePullingTest.java, https://github.com/data-integrations/google-cloud/blob/develop/src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiverSubscriptionTest.java

The team is looking into writing a unit test for this change. Considering the priority of this customer bug, can we take up the addition of a JUnit as a separate activity? We can use an existing issue - [PLUGIN-1399](https://cdap.atlassian.net/browse/PLUGIN-1399) or I can create a new issue on the PLUGIN JIRA Board and we can track its progress from there.

rmstar (approved):
Please squash commits after testing changes

sgarg-CS:
> Please squash commits after testing changes

We tested out the changes we made recently, and the pipeline was behaving as expected.
I've squashed the commits. Thank you!

PR #153, opened 2023-01-27
Fix for Salesforce plugin validation timeout

rmstar:
- Bump up timeout for getting oauth info to 2 mins
- Don't throw Runtime exception during plugin validation

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
+++ b/src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
@@ -18,45 +18,54 @@ package io.cdap.plugin.salesforce.authenticator;
 
 import com.google.common.base.Strings;
 import com.google.gson.Gson;
+import com.sforce.ws.ConnectionException;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
```diff
@@ -18,45 +18,54 @@
...
-      connectorConfig.setRestEndpoint(restEndpoint);
-      connectorConfig.setServiceEndpoint(serviceEndPoint);
-      // This should only be false when doing debugging.
-      connectorConfig.setCompression(true);
-      // Set this to true to see HTTP requests and responses on stdout
-      connectorConfig.setTraceMessage(false);
-
-      return connectorConfig;
+      oAuthInfo = getOAuthInfo(credentials);
     } catch (Exception e) {
```
rmstar, line 52: TODO: figure out if we can differentiate between different failure causes. e.g: connection timed out vs permission denied etc.

PR #155, opened 2023-01-28
Test

rmstar:


src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
```diff
@@ -38,8 +38,10 @@ public class Authenticator {
    *
    * @return ConnectorConfig which can be used to create BulkConnection and PartnerConnection
    */
-  public static ConnectorConfig createConnectorConfig(AuthenticatorCredentials credentials) {
-    try {
+  public static ConnectorConfig createConnectorConfig(AuthenticatorCredentials credentials) throws ConnectionException {
+    throw new ConnectionException("oops");
```
rmstar, line 42: This should show up in the task workers

PR #156, opened 2023-01-28
Improve the exception error message to better capture underlying issues.

wyzhang:
Why:
Existing exception error messages often don't include any udnerlying issues, and they are not surfaced to user. For instance, some users and customers experienced error of "There was issue communicating with Salesforce" without any more details provided.


greeshmaswaminathan (approved):

greeshmaswaminathan (changes requested):
Removing the approval because we are expecting more commits.

PR #157, opened 2023-01-31
Bug/cvs bug cs

vikasrathee-cs:
PR for testing

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
@@ -23,6 +23,7 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
 import io.cdap.plugin.salesforce.SalesforceConstants;
+import io.cdap.plugin.salesforce.authenticator.Authenticator;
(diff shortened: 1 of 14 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
```diff
@@ -141,9 +142,14 @@ public AuthenticatorCredentials getAuthenticatorCredentials() {
     if (oAuthInfo != null) {
       return new AuthenticatorCredentials(oAuthInfo);
     }
-
-    return new AuthenticatorCredentials(getUsername(), getPassword(), getConsumerKey(),
-                                        getConsumerSecret(), getLoginUrl());
+    try {
+      this.oAuthInfo = Authenticator.getOAuthInfo(new AuthenticatorCredentials(getUsername(), getPassword(),
```
wyzhang, line 146: what's the expiration for this token? will that be an issue if we "cache" the token

vikasrathee-cs: There is no expiration time for the token unless session is active. I am pasting a snippet from Salesforce discussions page below for more clarity. And even before also, we were getting the accessToken once and same was passed to RecordReader class to make further API calls.

Sessions expire based on your organization's policy for sessions. Basically, as long as the app is in active use, the session won't expire. Once the session is logged out, the timeout has elapsed, or it is otherwise expired (e.g. an administrator expires all sessions for the Connected App).

PR #158, opened 2023-01-31
Improve the exception error message to better capture underlying issues.

wyzhang:
Why:
Existing exception error messages often don't include any udnerlying issues, and they are not surfaced to user. For instance, some users and customers experienced error of "There was issue communicating with Salesforce" without any more details provided.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorConfig.java
@@ -181,7 +181,10 @@ public class SalesforceConnectorConfig extends PluginConfig {
     try {
       SalesforceConnectionUtil.getPartnerConnection(this.getAuthenticatorCredentials());
     } catch (ConnectionException e) {
-      throw new RuntimeException("There was issue communicating with Salesforce. " + e.getMessage(), e);
+      throw new RuntimeException(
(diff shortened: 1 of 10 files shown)


yaojiefeng (approved):
Nit - in cdap, the indentation is 2 characters instead of 4

greeshmaswaminathan (approved):

wyzhang:
Fixed. Imported and use cask style

src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorConfig.java
```diff
@@ -181,7 +181,10 @@ private void validateConnection() {
     try {
       SalesforceConnectionUtil.getPartnerConnection(this.getAuthenticatorCredentials());
     } catch (ConnectionException e) {
-      throw new RuntimeException("There was issue communicating with Salesforce. " + e.getMessage(), e);
+      throw new RuntimeException(
```
vikasrathee-cs, line 184: For ConnectionException, it is not returning any value in e.getMessage.
Actual error message can be called using its child class's method e.getExceptionMessage(), most of the classes which extend ConnectionException implements the IApiFault interface which has these methods. There we can get a proper message sent by Salesforce APIs.

vikasrathee-cs:
https://github.com/data-integrations/salesforce/pull/166
PR already merged.

PR #176, opened 2023-03-29
Proxy Support Added in all 4 plugins for release 1.4.

vikasrathee-cs:
Proxy Support Added in all 4 plugins. 3 new fields added in UI in connector and all 4 plugins.
https://cdap.atlassian.net/browse/PLUGIN-1538

### DIFF
--- a/docs/Salesforce-batchsink.md
+++ b/docs/Salesforce-batchsink.md
@@ -30,6 +30,8 @@ A Salesforce connected application must be created in order to get a client secr
 
 **Connect Timeout:** Maximum time in milliseconds to wait for connection initialization before time out.
 
+**Proxy URL:** Proxy URL. Must contain a protocol, address and port.
+
 **SObject Name:** Salesforce object name to insert records into.
 
 There are also **sObjects** that are not supported in the Bulk API of Salesforce.
@@ -72,17 +74,17 @@ Below is a non-comprehensive list of **sObjects** that are not currently availab
 - WorkOrderLineItemStatus
 - WorkOrderStatus
 
-**Operation:** Operation used for writing data into Salesforce.<br>
-Insert - adds records.<br>
+**Operation:** Operation used for writing data into Salesforce.  
+Insert - adds records.  
 Upsert - upserts the records. Salesforce will decide if sObjects 
-are the same using external id field.<br>
+are the same using external id field.  
 Update - updates existing records based on Id field.
 
 **Upsert External ID Field:** External id field name. It is used only if operation is upsert.
 The field specified can be either 'Id' or any customly created field, which has external id attribute set.
 
-**Concurrency Mode:** The concurrency mode for the bulk job. The valid values are:<br>
-Parallel - Process batches in parallel mode. This is the default value.<br>
+**Concurrency Mode:** The concurrency mode for the bulk job. The valid values are:  
+Parallel - Process batches in parallel mode. This is the default value.  
 Serial - Process batches in serial mode. Processing in parallel can cause database contention.
 When this is severe, the job can fail. If you’re experiencing this issue, submit the job with serial 
 concurrency mode. This mode guarantees that batches are processed one at a time, but can significantly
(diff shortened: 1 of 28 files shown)


docs/Salesforce-batchsink.md
```diff
@@ -30,6 +30,8 @@ A Salesforce connected application must be created in order to get a client secr
 
 **Connect Timeout:** Maximum time in milliseconds to wait for connection initialization before time out.
 
+**Proxy URL:** Proxy URL. Must contain a protocol, address and port.
```
greeshmaswaminathan, line 33: Can the other changes be reverted in this md file? Looks like formatting changes?

docs/Salesforce-batchsource.md
```diff
@@ -31,6 +27,12 @@ A Salesforce connected application must be created in order to get a client secr
 
 **Connect Timeout:** Maximum time in milliseconds to wait for connection initialization before time out.
 
+**Proxy URL:** Proxy URL. Must contain a protocol, address and port.
+
+**Username:** Proxy username.
```
greeshmaswaminathan: Proxy username and password is not applicable, correct?

src/main/java/io/cdap/plugin/salesforce/authenticator/AuthenticatorCredentials.java
```diff
@@ -97,6 +101,11 @@ public Integer getConnectTimeout() {
     return connectTimeout;
   }
 
+  @Nullable
+  public String getProxyUrl() {
+    return proxyUrl;
```
greeshmaswaminathan, line 106: Can you check if the new fields introduced need to be part of the equals and hashcode ?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
```diff
@@ -100,16 +100,18 @@ protected SalesforceBaseSourceConfig(String referenceName,
...
-          oAuthInfo);
+          oAuthInfo, proxyUrl);
     this.datetimeAfter = datetimeAfter;
     this.datetimeBefore = datetimeBefore;
     this.duration = duration;
     this.offset = offset;
     this.operation = operation;
   }
 
+
```
greeshmaswaminathan: nit: remove the new line.

docs/Salesforce-batchsink.md
vikasrathee-cs: other changes are related to removing <br> tag and use double space instead. We can keep those changes.

docs/Salesforce-batchsource.md
vikasrathee-cs: removed.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBaseSourceConfig.java
vikasrathee-cs: removed

src/main/java/io/cdap/plugin/salesforce/authenticator/AuthenticatorCredentials.java
vikasrathee-cs: New fields are not required to be part of hashcode or equals method as new fields are non mandatory and we are doing any comparison of AuthenticatorCredentials anywhere or storing it into any hash data structure. We can skip that.

greeshmaswaminathan:
@vikasrathee-cs , Can you raise a separate PR for https://cdap.atlassian.net/browse/CDAP-20537 ?

src/main/java/io/cdap/plugin/salesforce/authenticator/AuthenticatorCredentials.java
greeshmaswaminathan: Suggestion: Even if they are optional, it would make a difference to the object equality. I agree that it is not relevant to this particular class since it may not be used in hash data structures. However since they are already defined, good to add the new fields to keep it correct.

vikasrathee-cs: added.

greeshmaswaminathan (approved):

PR #179, opened 2023-04-05
bugfix for CDAP-20537 - release 1.4 branch

vikasrathee-cs:
Seeing null input schema error when using Salesforce plugin v 1.4.4
https://cdap.atlassian.net/browse/CDAP-20537

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
@@ -149,7 +149,7 @@ public class BaseSalesforceConfig extends ReferencePluginConfig {
     return connectTimeout;
   }
 
-  public void validate(FailureCollector collector, OAuthInfo oAuthInfo) {
+  public void validate(FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
     try {
       validateConnection(oAuthInfo);
     } catch (Exception e) {
@@ -197,8 +197,8 @@ public class BaseSalesforceConfig extends ReferencePluginConfig {
       || containsMacro(SalesforceConstants.PROPERTY_CONNECT_TIMEOUT));
   }
 
-  private void validateConnection(OAuthInfo oAuthInfo) {
-    if (!canAttemptToEstablishConnection()) {
+  private void validateConnection(@Nullable OAuthInfo oAuthInfo) {
+    if (oAuthInfo == null) {
       return;
     }
 
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
@@ -203,7 +203,7 @@ public class SalesforceSinkConfig extends BaseSalesforceConfig {
                                                     SalesforceSinkConfig.PROPERTY_ERROR_HANDLING));
   }
 
-  public void validate(Schema schema, FailureCollector collector, OAuthInfo oAuthInfo) {
+  public void validate(Schema schema, FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
     super.validate(collector, oAuthInfo);
     validateSinkProperties(collector);
     validateSchema(schema, collector, oAuthInfo);
@@ -265,14 +265,13 @@ public class SalesforceSinkConfig extends BaseSalesforceConfig {
     collector.getOrThrowException();
   }
 
-  private void validateSchema(Schema schema, FailureCollector collector, OAuthInfo oAuthInfo) {
-    List<Schema.Field> fields = schema.getFields();
-    if (fields == null || fields.isEmpty()) {
+  private void validateSchema(Schema schema, FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
+    if (schema == null || schema.getFields() == null || schema.getFields().isEmpty()) {
       collector.addFailure("Sink schema must contain at least one field", null);
       throw collector.getOrThrowException();
     }
 
-    if (!canAttemptToEstablishConnection() || containsMacro(PROPERTY_SOBJECT)
+    if (oAuthInfo == null || containsMacro(PROPERTY_SOBJECT)
       || containsMacro(PROPERTY_OPERATION) || containsMacro(PROPERTY_EXTERNAL_ID_FIELD)) {
       return;
     }
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
@@ -73,12 +73,8 @@ public class SalesforceBatchSource extends BatchSource<Schema, Map<String, Strin
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
-    // validate when macros not yet substituted
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
-    if (!config.canAttemptToEstablishConnection()) {
-      config.validateSOQLQueryParameters(collector);
-      return;
-    }
+
     OAuthInfo oAuthInfo = SalesforceConnectionUtil.getOAuthInfo(config, collector);
     config.validate(collector, oAuthInfo);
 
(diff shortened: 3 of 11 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
greeshmaswaminathan, line 1: Can you also check if the schema is set correctly for multi source and streaming source plugins?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -77,6 +77,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     if (!config.canAttemptToEstablishConnection()) {
```
greeshmaswaminathan: Suggestion: Can we consolidate the logic better in this method? There is also a check further down for `!config.canAttemptToEstablishConnection()` which will never be true. The config has its own macro checks as well.

greeshmaswaminathan:
Pleas raise a PR against develop as well.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
albertshau: +1

please also add a unit test

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
vikasrathee-cs: For multisource schema will always be null, for streaming plugin also issue is fixed and added to the PR.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
vikasrathee-cs: unnecessary checks removed for all plugins and unit tests added for this.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -73,12 +73,14 @@ public SalesforceBatchSource(SalesforceSourceConfig config) {
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
-    // validate when macros not yet substituted
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    // validate other filter fields when connection is not available and set schema from config
     if (!config.canAttemptToEstablishConnection()) {
       config.validateSOQLQueryParameters(collector);
```
greeshmaswaminathan: It is better to check for `canAttemptToEstablishConnection` while getting oAuthInfo  and mark it nullable where ever it is used.  The method `validateSOQLQueryParameters` can be removed since it is mostly duplicate code. Let config handle all the validations in `validate`.

src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
```diff
@@ -244,4 +250,35 @@ private Field getFieldWithType(String name, FieldType type, boolean isNillable)
...
   }
+  
+  @Test
+  public void testSourceSchemaNotNullIfConnectionMacroAndImportManually() {
+    Schema schema = Schema.recordOf("output",
+                                    Schema.Field.of("Id",
+                                                    Schema.of(Schema.Type.STRING)));
+    SalesforceSourceConfig mockConfig = Mockito.mock(SalesforceSourceConfig.class);
+    MockPipelineConfigurer mockPipelineConfigurer = new MockPipelineConfigurer(null);
+    SalesforceBatchSource source = new SalesforceBatchSource(mockConfig);
```
greeshmaswaminathan, line 263: Please add a test for multi source.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
vikasrathee-cs: done. canAttemptToEstablishConnection check is added before using OAuthInfo.

src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
vikasrathee-cs: Added. But for Multisource, output schema will always be null.

src/main/java/io/cdap/plugin/salesforce/plugin/BaseSalesforceConfig.java
```diff
@@ -197,7 +197,7 @@ public boolean canAttemptToEstablishConnection() {
       || containsMacro(SalesforceConstants.PROPERTY_CONNECT_TIMEOUT));
   }
 
-  private void validateConnection(OAuthInfo oAuthInfo) {
+  private void validateConnection(@Nullable OAuthInfo oAuthInfo) {
     if (!canAttemptToEstablishConnection()) {
```
greeshmaswaminathan: The `canAttemptToEstablishConnection` check here can be replaced with a null check, since it is already checked while attempting to get OAuthInfo.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -265,9 +265,8 @@ public void validateSinkProperties(FailureCollector collector) {
     collector.getOrThrowException();
   }
 
-  private void validateSchema(Schema schema, FailureCollector collector, OAuthInfo oAuthInfo) {
-    List<Schema.Field> fields = schema.getFields();
-    if (fields == null || fields.isEmpty()) {
+  private void validateSchema(Schema schema, FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
```
greeshmaswaminathan, line 268: Use null checks where oAuthInfo is being used.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -73,12 +73,8 @@ public SalesforceBatchSource(SalesforceSourceConfig config) {
...
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
-    // validate when macros not yet substituted
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
-    if (!config.canAttemptToEstablishConnection()) {
-      config.validateSOQLQueryParameters(collector);
-      return;
-    }
+
     OAuthInfo oAuthInfo = SalesforceConnectionUtil.getOAuthInfo(config, collector);
```
greeshmaswaminathan, line 78: Include null check for `OAuthInfo` since it is nullable.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceMultiSourceConfig.java
```diff
@@ -162,7 +162,10 @@ public List<String> getQueries(long logicalStartTime, OAuthInfo oAuthInfo) {
   /**
    * validate whiteList and blackList SObjects
    */
-  public void validateSObjects(FailureCollector collector, OAuthInfo oAuthInfo) {
+  public void validateSObjects(FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
+    if (!canAttemptToEstablishConnection()) {
```
greeshmaswaminathan: Use a null check on OAuthInfo instead. That is more obvious in this context.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -164,7 +164,7 @@ public boolean isSoqlQuery() {
                                      SalesforceSourceConstants.PROPERTY_QUERY);
   }
 
-  public void validate(FailureCollector collector, OAuthInfo oAuthInfo) {
+  public void validate(FailureCollector collector, @Nullable OAuthInfo oAuthInfo) {
```
greeshmaswaminathan, line 167: Use null checks for `OAuthInfo`.

greeshmaswaminathan (approved):

PR #187, opened 2023-05-09
Added geolocation in supported datatypes

Aryan-Verma:
Added geolocation (latitude) and geolocation (longitude) in the Salesforce to CDAP datatype mapping table documentation as part of datatype handling consistency investigation.


MrRahulSharma (approved):
LGTM

PR #207, opened 2023-07-11
e2e-salesforce-datavalidation

AnkitCLI:


src/e2e-test/java/io/cdap/plugin/BQValidation.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  public static boolean validateSalesforceMultiObjectToBQRecordValues() throws IOException, InterruptedException {
+    List<String> objectNames = Arrays.asList(SObjects.AUTOMATION_CUSTOM__C.value, SObjects.AUTOMATION_CUSTOM2__C.value);
+    List<String> targetTables = getTableByName();
+    if (objectNames.size() != targetTables.size()) {
+      throw new IllegalArgumentException("Number of object names and target tables must be the same.");
+    }
+
+    for (int i = 0; i < objectNames.size(); i++) {
```
Aryan-Verma: please use a different name for the variable, like currentObject etc

src/e2e-test/java/io/cdap/plugin/BQValidation.java
```diff
@@ -0,0 +1,267 @@
...
+            Date sourceDate = (jsonElementSource != null && !jsonElementSource.isJsonNull()) ? Date.valueOf(
+              jsonElementSource.getAsString()) : null;
+            JsonElement jsonElementTarget = bigQueryData.get(jsonObjectIdx).get(columnName);
+            Date targetDate = (jsonElementTarget != null && !jsonElementTarget.isJsonNull()) ? Date.valueOf(
+              jsonElementTarget.getAsString()) : null;
+            Assert.assertEquals("Different values found for column : %s", sourceDate, targetDate);
+            break;
+
+          default:
+
```
Aryan-Verma: nit: remove empty line

src/e2e-test/java/io/cdap/plugin/BQValidation.java
AnkitCLI: done

src/e2e-test/java/io/cdap/plugin/BQValidation.java
AnkitCLI: done

src/e2e-test/java/io/cdap/plugin/BQValidation.java
saimukkamala: nit: instead of using `i` as a variable name, can we use something meaningful like `objNameIndex`

AnkitCLI: done @b164706e5fce61a001b59979c1a2e73e5b70fea7

Aryan-Verma (approved):

src/e2e-test/java/io/cdap/plugin/BQValidation.java
```diff
@@ -0,0 +1,260 @@
...
+    for (Object rows : bigQueryRows) {
+      JsonObject jsonData = gson.fromJson(String.valueOf(rows), JsonObject.class);
+      bigQueryResponse.add(jsonData);
+    }
+    List<JsonObject> sObjectResponse;
+    sObjectResponse = SalesforceClient.queryObject(uniqueRecordId, objectName);
+    return compareSalesforceAndJsonData(sObjectResponse, bigQueryResponse, targetTable);
+  }
+
+  public static boolean validateBQToSalesforceRecordValues(String targetTable, String objectName) throws
```
saimukkamala: `validateBQToSalesforceRecordValues`, `validateSalesforceToBQRecordValues` - both functions seem to the exact same thing. Is there a reason why we can't merge or move the functionality to a single function?

saimukkamala (approved):

src/e2e-test/java/io/cdap/plugin/BQValidation.java
AnkitCLI: These two functions are used in two different steps here. The first is for SFtoBQ, in which we create a target table in the sink, and the second is for BQtoSF, in which we create a source table using hooks. As a result, the parameters in both functions that we provide are different.

saimukkamala: I see, the order of input parameters is different to indicate whether we are moving data from bq to salesforce or otherwise. But the functionality is exactly the same inside the function. and this functionality can be moved to a common function and and both `validateBQToSalesforceRecordValues` and `validateSalesforceToBQRecordValues` can call this function internally.

AnkitCLI: Done @ce261081c0bce6b72e3fb955ef9b6c2c0c6b1087

Vipinofficial11:
Merging the PR as the comments are addressed  and the build is green.

PR #209, opened 2023-07-14
Salesforce refactoring work

MrRahulSharma:
Salesforce changes for CS integration

Pending/Ongoing work:
1. Addition of Unit Tests
2. Changing the RecordReader logic to use the RecordReaderWrapper.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
@@ -32,7 +32,6 @@ public class SalesforceQueryUtil {
   private static final String WHERE = " WHERE ";
   private static final String AND = " AND ";
 
-
   private static final String FIELD_LAST_MODIFIED_DATE = "LastModifiedDate";
   private static final String FIELD_ID = "Id";
 
--- a/src/main/java/io/cdap/plugin/salesforce/connector/SalesforceConnector.java
+++ b/src/main/java/io/cdap/plugin/salesforce/connector/SalesforceConnector.java
@@ -74,7 +74,7 @@ public class SalesforceConnector implements DirectConnector {
   private final SalesforceConnectorConfig config;
   private StructuredRecord record;
 
-  SalesforceConnector(SalesforceConnectorConfig config) {
+  public SalesforceConnector(SalesforceConnectorConfig config) {
     this.config = config;
   }
 
@@ -87,6 +87,16 @@ public class SalesforceConnector implements DirectConnector {
 
   @Override
   public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest browseRequest) throws IOException {
+    return browse();
+  }
+
+  /**
+   * Browse functionality based on the config.
+   *
+   * @return BrowseDetail for the given config
+   * @throws IOException In case of Salesforce connection failure while browsing
+   */
+  public BrowseDetail browse() throws IOException {
     AuthenticatorCredentials credentials = new AuthenticatorCredentials(config.getUsername(), config.getPassword(),
                                                                         config.getConsumerKey(),
                                                                         config.getConsumerSecret(),
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
@@ -47,6 +47,7 @@ import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
 import io.cdap.plugin.salesforce.plugin.OAuthInfo;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSourceConstants;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSplitUtil;
+import org.apache.hadoop.mapreduce.RecordReader;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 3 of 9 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -74,7 +76,20 @@ public RecordReader createRecordReader(InputSplit split, TaskAttemptContext cont
     return new SalesforceRecordReaderWrapper(sObjectName, sObjectNameField, getDelegateRecordReader(query, schema));
   }
 
-  private RecordReader<Schema, Map<String, ?>> getDelegateRecordReader(String query, Schema schema) {
+  public static RecordReader createInitedRecordReader(
```
fernst: "Inited" sounds weird. Do you mean `initialized`? Why not call this `createRecordReader`?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceInputFormat.java
```diff
@@ -88,4 +103,18 @@ public RecordReader createRecordReader(InputSplit split, TaskAttemptContext cont
     return new SalesforceWideRecordReader(schema, query, new SoapRecordToMapTransformer());
   }
 
+  private static RecordReader<Schema, Map<String, ?>> getInitedDelegateRecordReader(
```
fernst: Same comment as before. `inited` sounds weird.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceSourceConfig.java
```diff
@@ -90,7 +90,7 @@ public class SalesforceSourceConfig extends SalesforceBaseSourceConfig {
   private final String parent;
 
   @VisibleForTesting
-  SalesforceSourceConfig(String referenceName,
+  public SalesforceSourceConfig(String referenceName,
```
fernst, line 92: if this function is public for more than just being visible for testing, we should remove the `@VisibleForTesting` annotation

fernst (approved):
Code LGTM overall, left some comments/suggestions

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
```diff
@@ -50,6 +50,7 @@ public static String createSObjectQuery(List<String> fields, String sObjectName,
       .append(SELECT).append(String.join(",", fields))
       .append(FROM).append(sObjectName);
 
+
```
tivv: nit: double newline

src/main/java/io/cdap/plugin/salesforce/connector/SalesforceConnector.java
```diff
@@ -85,8 +85,13 @@ public void test(ConnectorContext connectorContext) throws ValidationException {
     config.validate(collector, oAuthInfo);
   }
 
+
```
tivv: nit: double newline

src/main/java/io/cdap/plugin/salesforce/connector/SalesforceConnector.java
```diff
@@ -85,8 +85,13 @@ public void test(ConnectorContext connectorContext) throws ValidationException {
...
   @Override
   public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest browseRequest) throws IOException {
+    return browse(this.config);
+  }
+
+  public BrowseDetail browse(SalesforceConnectorConfig config) throws IOException {
```
tivv: Why do you need to expose both this and constructor? Usually you either pass something in constructor or as method parameter, but not both.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -179,25 +194,43 @@ public void transform(KeyValue<Schema, Map<String, String>> input,
...
+   * Get the RecordReader for a given split.
+   *
+   * @param salesforceSplit Given Salesforce Split
+   * @param schemaJsonString Schema JSON
+   * @param credentials Credential object
+   * @return RecordReader for the given split
+   */
+  public static RecordReader<Schema, Map<String, ?>> getRecordReader (SalesforceSplit salesforceSplit,
+                                                                      String schemaJsonString,
+                                                                      AuthenticatorCredentials credentials) {
```
tivv: Why do we need this method in this class? Can caller just call the createRecordReader directly?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -122,36 +125,48 @@ public void prepareRun(BatchSourceContext context) {
...
-                                 .map(Schema.Field::getName)
-                                 .collect(Collectors.toList()));
+        Preconditions.checkNotNull(schema.getFields()).stream()
+            .map(Schema.Field::getName)
+            .collect(Collectors.toList()));
 
     authenticatorCredentials = config.getConnection().getAuthenticatorCredentials();
+    List<SalesforceSplit> querySplits =
+        getSplits(config, authenticatorCredentials, context.getLogicalStartTime(), oAuthInfo);
+    querySplits.parallelStream().forEach(salesforceSplit -> jobIds.add(salesforceSplit.getJobId()));
```
tivv: jobIds is not thread-safe, so please don't use parallelism. Why do you need parallelism here?

MrRahulSharma: This is coming as part of the refactoring of the existing code, since it was already using parallel stream, hence kept it the same here.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
MrRahulSharma: I believe we can.

src/main/java/io/cdap/plugin/salesforce/connector/SalesforceConnector.java
MrRahulSharma: done.

tivv (approved):

PR #212, opened 2023-08-02
PLUGIN-1656 Added a field for datatype validation to avoid validation.

vikasrathee-cs:
https://cdap.atlassian.net/browse/PLUGIN-1656
Added a field for datatype validation to avoid validation in Salesforce sink plugin. E.g. some existing users are providing date fields as a string in input schema and same was not getting validated and this process was working fine as while creating csv file these string were getting used as it as.  We added validation on data types to avoid any wrong value in such fields but customer wants to use existing approach. So, we have added a flag datatype validation which will be false by default so that it will not impact existing users.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
@@ -178,8 +178,8 @@ public class SalesforceSchemaUtil {
       if (!actualFieldSchema.equals(providedFieldSchema)
         || !Objects.equals(actualFieldSchema.getLogicalType(), providedFieldSchema.getLogicalType())) {
         throw new IllegalArgumentException(
-          String.format("Expected field '%s' to be of '%s', but it is of '%s'",
-            providedField.getName(), providedFieldSchema, actualFieldSchema));
+          String.format("Expected field '%s' to be of %s, but it is of %s",
+                        providedField.getName(), providedFieldSchema, actualFieldSchema));
       }
 
       if (checkNullable && isActualFieldNullable && !isProvidedFieldNullable) {
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
@@ -58,6 +58,7 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
   public static final String PROPERTY_OPERATION = "operation";
   public static final String PROPERTY_EXTERNAL_ID_FIELD = "externalIdField";
   public static final String PROPERTY_CONCURRENCY_MODE = "concurrencyMode";
+  public static final String PROPERTY_DATATYPE_VALIDATION = "datatypeValidation";
 
   private static final String SALESFORCE_ID_FIELD = "Id";
 
@@ -133,6 +134,12 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
   @Description("The existing connection to use.")
   private SalesforceConnectorConfig connection;
 
+  @Name(PROPERTY_DATATYPE_VALIDATION)
+  @Nullable
+  @Macro
+  @Description("Whether to validate the field data types of the input schema as per Salesforce specific data types")
+  private final Boolean datatypeValidation;
+
   public SalesforceSinkConfig(String referenceName,
                               @Nullable String clientId,
                               @Nullable String clientSecret,
@@ -146,7 +153,8 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
                               String errorHandling,
                               @Nullable String securityToken,
                               @Nullable OAuthInfo oAuthInfo,
-                              @Nullable String proxyUrl) {
+                              @Nullable String proxyUrl,
+                              @Nullable Boolean datatypeValidation) {
     super(referenceName);
     connection = new SalesforceConnectorConfig(clientId, clientSecret, username, password, loginUrl,
                                                securityToken, connectTimeout, oAuthInfo, proxyUrl);
@@ -157,6 +165,7 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
     this.maxBytesPerBatch = maxBytesPerBatch;
     this.maxRecordsPerBatch = maxRecordsPerBatch;
     this.errorHandling = errorHandling;
+    this.datatypeValidation = datatypeValidation;
   }
 
   private static final String DEFAULT_LOGIN_URL = "https://login.salesforce.com/services/oauth2/token";
@@ -188,7 +197,7 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
   }
 
   public String getConcurrencyMode() {
-    return concurrencyMode;
+    return concurrencyMode == null ? ConcurrencyMode.Parallel.name() : concurrencyMode;
   }
 
   public ConcurrencyMode getConcurrencyModeEnum() {
@@ -220,8 +229,11 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
 
   public ErrorHandling getErrorHandling() {
     return ErrorHandling.fromValue(errorHandling)
-      .orElseThrow(() -> new InvalidConfigException("Unsupported error handling value: " + errorHandling,
-                                                    SalesforceSinkConfig.PROPERTY_ERROR_HANDLING));
+      .orElse(ErrorHandling.FAIL);
+  }
+
+  public boolean isDatatypeValidation() {
+    return datatypeValidation != null && datatypeValidation;
   }
 
   public String getReferenceNameOrNormalizedFQN(String orgId, String sObject) {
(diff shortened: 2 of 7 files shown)


MrRahulSharma:
Is this a regression ?
Can you update the JIRA and share failure stacktrace ?

vikasrathee-cs:
> Is this a regression ? Can you update the JIRA and share failure stacktrace ?

Yes, it was added to validate datatypes also along with name in the schema but previous users don't want this validation as plugin was working fine even if it get mapped with wrong data type. Added a comment in the JIRA for exact error that customer is facing.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -146,7 +153,8 @@ public SalesforceSinkConfig(String referenceName,
                               String errorHandling,
                               @Nullable String securityToken,
                               @Nullable OAuthInfo oAuthInfo,
-                              @Nullable String proxyUrl) {
+                              @Nullable String proxyUrl,
+                              @Nullable boolean datatypeValidation) {
```
tivv: Nullable does not apply for primitive types. Did you mean it to be "Boolean"?

widgets/Salesforce-batchsink.json
```diff
@@ -193,6 +193,22 @@
...
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "YES"
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "false"
```
tivv: Should we have it "true" here to ensure any new pipeline has it enabled?

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
vikasrathee-cs: will make it Boolean

src/test/java/io/cdap/plugin/salesforce/connector/SalesforceConnectorTest.java
```diff
@@ -96,6 +98,33 @@ private Schema getPluginSchema() throws IOException {
...
+    SalesforceSinkConfig config = new SalesforceSinkConfig("REFERENCE_NAME",
+                                                           "CONSUMER_KEY", "CONSUMER_SECRET",
+                                                           "USERNAME", "PASSWORD",
+                                                           "LOGIN_URL", 30000, "sObject", "Insert", null,
+                                                           ConcurrencyMode.Parallel.name(), "1000000", "10000",
+                                                           "Fail on Error",
+                                                           "SECURITY_TOKEN",
+                                                           null, null, true);
 
+    if (config.isDatatypeValidation()) {
```
rmstar: The test should not call isDatatypeValidation() directly - call the validate method and verify in the test that input schema was validated.
Similarly, add another test for the case where datatypeValidation=false and verify in the test that we don't do schema validation

widgets/Salesforce-batchsink.json
vikasrathee-cs: Will make true and in case of upgrading a pipeline if this value is not available, will make it false

rmstar: +1 please test this case manually as well. 
1. Pipeline upgrade: salesforce sink v1.4.7 -> v1.4.8 - data type validation should be disabled
2. Create a new pipeline with salesforce sink v1.4.8 with default settings for data type validation => data type validation should be enabled

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
```diff
@@ -133,6 +134,12 @@ public class SalesforceSinkConfig extends ReferencePluginConfig {
   @Description("The existing connection to use.")
   private SalesforceConnectorConfig connection;
 
+  @Name(PROPERTY_DATATYPE_VALIDATION)
+  @Nullable
+  @Macro
```
rmstar, line 139: do you need a containsMacro() check for this field?

widgets/Salesforce-batchsink.json
```diff
@@ -193,6 +193,22 @@
             ],
             "default": "Skip on error"
           }
+        },
+        {
+          "widget-type": "toggle",
+          "label": "Datatype Validation",
```
rmstar: nit: Data type (add space).

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceSinkConfig.java
vikasrathee-cs: will add that.

src/test/java/io/cdap/plugin/salesforce/connector/SalesforceConnectorTest.java
vikasrathee-cs: I have tried that but the checkCompatibility method inside validateInputSchema  is returning void and its class is mocked static, so we need to upgrade Mockito library for that and upgrading that causing various test failures in existing test cases. As this is a P1 issue, we tried to expedite it, and that will require more time. Can we add a ETL test case for time being to test it fully.

rmstar: Sure, we can update tests later. You can create a Jira for tracking purposes and update the tests in another PR

widgets/Salesforce-batchsink.json
vikasrathee-cs: This test case is tested.

pom.xml
```diff
@@ -22,7 +22,7 @@
   <name>Salesforce plugins</name>
   <groupId>io.cdap.plugin</groupId>
   <artifactId>salesforce-plugins</artifactId>
-  <version>1.6.0-SNAPSHOT</version>
+  <version>1.7.0-SNAPSHOT</version>
   <packaging>jar</packaging>
```
MrRahulSharma: We do not require to do this change here.

src/main/java/io/cdap/plugin/salesforce/SalesforceSchemaUtil.java
```diff
@@ -178,8 +178,8 @@ public static void checkCompatibility(Schema actualSchema, Schema providedSchema
       if (!actualFieldSchema.equals(providedFieldSchema)
         || !Objects.equals(actualFieldSchema.getLogicalType(), providedFieldSchema.getLogicalType())) {
         throw new IllegalArgumentException(
-          String.format("Expected field '%s' to be of '%s', but it is of '%s'",
-            providedField.getName(), providedFieldSchema, actualFieldSchema));
+          String.format("Expected field '%s' to be of %s, but it is of %s",
+                        providedField.getName(), providedFieldSchema, actualFieldSchema));
```
MrRahulSharma, line 182: Why do we need this change ?

vikasrathee-cs: schema already have quotes in it, so it was coming like '"String"'.

src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSinkETLTest.java
```diff
@@ -231,6 +231,6 @@ protected SalesforceSinkConfig getDefaultConfig(String sObject) {
                                     BaseSalesforceETLTest.LOGIN_URL, 30000, sObject, "Insert", null,
                                     ConcurrencyMode.Parallel.name(), "1000000", "10000", "Fail on Error",
                                     BaseSalesforceETLTest.SECURITY_TOKEN,
-                                    null, null);
+                                    null, null, false);
```
MrRahulSharma, line 234: Please add a test for the true scenario as well.

widgets/Salesforce-batchsink.json
```diff
@@ -193,6 +193,22 @@
...
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "YES"
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "true"
```
MrRahulSharma, line 210: Shouldn't the default be false ?

src/test/java/io/cdap/plugin/salesforce/etl/BaseSalesforceBatchSinkETLTest.java
vikasrathee-cs: Unit test cases will be added separately as Mockito framework need some upgrade, separate ticket is added for that https://cdap.atlassian.net/browse/PLUGIN-1657

widgets/Salesforce-batchsink.json
vikasrathee-cs: it should be true for new pipelines and false for upgrade pipelines. In case of upgrade pipeline, we are setting it as false in isDatatypeValidation method.

pom.xml
vikasrathee-cs: removed.

tivv (approved):

PR #219, opened 2023-08-16
PLUGIN-1665 fix oauthInfo property

albertshau:
Fix usage of the oauthInfo plugin property. CDAP does not allow objects within objects in a PluginConfig. This was preventing the plugins from getting instantiated whenever the oauthInfo macro was being used.

Refactored classes to pull OAuthInfo into top level fields in each plugin config. Introduced a SalesforceConnectorInfo class that contains logic that used to be in the SalesforceConnectorConfig class, and changed the various plugin configs to return this new ConnectorInfo class instead of the ConnectorConfig class.

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorInfo.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorInfo.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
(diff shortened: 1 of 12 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorInfo.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
```
vikasrathee-cs: change the year for new class

src/main/java/io/cdap/plugin/salesforce/plugin/connector/SalesforceConnectorConfig.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
```
vikasrathee-cs, line 3: change year

src/main/java/io/cdap/plugin/salesforce/plugin/connector/SalesforceConnectorConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    super(consumerKey, consumerSecret, username, password, loginUrl, securityToken, connectTimeout, proxyUrl);
+    this.oAuthInfo = oAuthInfo;
+  }
+
+  @Nullable
+  public OAuthInfo getOAuthInfo() {
+    return oAuthInfo;
+  }
+
+  @Nullable
```
vikasrathee-cs: This should be part of SalesforceConnectorBaseConfig

vikasrathee-cs (approved):
LGTM. Squash the commits.

PR #236, opened 2024-01-02
Added code for aborting job in case of failure and logs added

vikasrathee-cs:
Added code for aborting job in case of failure and logs added


src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceRecordWriter.java
```diff
@@ -163,9 +163,18 @@ public void close(TaskAttemptContext taskAttemptContext) throws IOException {
       SalesforceBulkUtil.awaitCompletion(bulkConnection, jobInfo, batchInfoList,
                                          errorHandling.equals(ErrorHandling.SKIP));
       SalesforceBulkUtil.checkResults(bulkConnection, jobInfo, batchInfoList, errorHandling.equals(ErrorHandling.SKIP));
-    } catch (AsyncApiException | ConditionTimeoutException | BulkAPIBatchException e) {
+    } catch (AsyncApiException | ConditionTimeoutException e) {
       throw new RuntimeException(String.format("Failed to check the result of a batch for writes: %s",
                                                e.getMessage()), e);
+    } catch (BulkAPIBatchException e) {
```
MrRahulSharma, line 169: Please add details when does this exception is raised and what we behavior we intend to implement.

vikasrathee-cs: Added.

MrRahulSharma (approved):

PR #237, opened 2024-01-31
Added code to browse only the queryable sobjects.

MrRahulSharma:


src/main/java/io/cdap/plugin/salesforce/plugin/connector/SalesforceConnector.java
```diff
@@ -88,7 +88,7 @@ public void test(ConnectorContext connectorContext) throws ValidationException {
 
   @Override
   public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest browseRequest) throws IOException {
-    return browse();
+    return browse(false);
```
sumitjnn, line 91: Should queryableOnly be passed as true here to ensure only the queryable objects are returned?

MrRahulSharma: As per the current implementation, we do return all. Hence kept it false to avoid any regression.

src/main/java/io/cdap/plugin/salesforce/plugin/connector/SalesforceConnector.java
```diff
@@ -114,6 +114,13 @@ public BrowseDetail browse() throws IOException {
       for (int i = 0; i < dgr.getSobjects().length; i++) {
         String name = dgr.getSobjects()[i].getName();
         String label = dgr.getSobjects()[i].getLabel();
+        boolean queryable = dgr.getSobjects()[i].isQueryable();
+
+        // Continue in case of returning only queryable sObjects and the current sObject is non-queryable.
+        if (queryableOnly && !queryable) {
```
fernst: nit: change the input paramter name to `onlyReturnQueryableObjects` so it's more semantically self-explanatory.

Also, th `queryable` variable should be `isQueryable`

```
if (onlyReturnQueryableObjects && !isQueryable) { ... }
```

fernst (approved):

PR #246, opened 2024-05-14
Add option to allow null values in non-null custom fields in sObjects

vanathi-g:
## Changes
Since custom sObject fields that are marked as non-nullable can sometimes have null values, a new option is provided to allow these null values by marking the corresponding CDAP schema field as nullable.

### DIFF
--- a/src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
+++ b/src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
@@ -15,6 +15,8 @@
  */
 package io.cdap.plugin.salesforce;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.sforce.soap.partner.Field;
 import com.sforce.soap.partner.FieldType;
 import io.cdap.cdap.api.data.format.StructuredRecord;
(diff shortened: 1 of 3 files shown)


src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
```diff
@@ -166,6 +168,40 @@ public void testGetSchemaWithFields() {
...
+    objectFields.put("CreatedDate", getCustomFieldWithType("CreatedDate", FieldType.date, false));
+    SObjectsDescribeResult describeResult = SObjectsDescribeResult.of(ImmutableMap.of(objectName, objectFields));
+
+    Schema expectedSchema = Schema.recordOf("output",
+            Schema.Field.of("Name", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+            Schema.Field.of("Value", Schema.nullableOf(Schema.of(Schema.Type.DOUBLE))),
+            Schema.Field.of("CreatedDate", Schema.nullableOf(Schema.of(Schema.LogicalType.DATE)))
+    );
+
+    Schema actualSchema = SalesforceSchemaUtil.getSchemaWithFields(sObjectDescriptor, describeResult, true);
```
MrRahulSharma, line 201: Can you add another validation in this function with 'false' ?

MrRahulSharma (approved):
LGTM

src/test/java/io/cdap/plugin/salesforce/SalesforceSchemaUtilTest.java
vanathi-g: Done

PR #250, opened 2024-05-28
Add support for other access token grant types in Authenticator

vanathi-g:
## Changes
- Added a new "GrantType" enum and parameter to AuthenticatorCredentials
- Refactored AuthenticatorCredentials to infer the correct grant_type from a provided set of parameters
- Changed Authenticator to use the correct OAuth flow based on what grant_type is passed via AuthenticatorCredentials

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
+++ b/src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
@@ -25,6 +25,7 @@ import io.cdap.plugin.salesforce.plugin.OAuthInfo;
 import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.client.HttpProxy;
 import org.eclipse.jetty.client.ProxyConfiguration;
+import org.eclipse.jetty.client.api.Request;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
 import java.net.URI;
@@ -84,6 +85,10 @@ public class Authenticator {
       return oAuthInfo;
     }
 
+    if (credentials.getGrantType() == null) {
+      throw new IllegalArgumentException("Grant type cannot be null for OAuth flow to fetch access token.");
+    }
+
     SslContextFactory sslContextFactory = new SslContextFactory();
     HttpClient httpClient = new HttpClient(sslContextFactory);
     httpClient.setConnectTimeout(credentials.getConnectTimeout());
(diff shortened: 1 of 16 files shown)


src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
```diff
@@ -92,12 +97,24 @@ public static OAuthInfo getOAuthInfo(AuthenticatorCredentials credentials) throw
...
+      Request authRequest = httpClient.POST(credentials.getLoginUrl())
+              .param("grant_type", credentials.getGrantType().getType());
+
+      switch(credentials.getGrantType()) {
+        case CLIENT_CREDENTIALS:
+          authRequest = authRequest.param("client_id", credentials.getConsumerKey())
+                  .param("client_secret", credentials.getConsumerSecret());
+          break;
+        case PASSWORD:
+          authRequest = authRequest.param("grant_type", "password")
```
vikasrathee-cs: grant type already set in line 102.

src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
```diff
@@ -84,6 +85,10 @@ public static OAuthInfo getOAuthInfo(AuthenticatorCredentials credentials) throw
       return oAuthInfo;
     }
 
+    if (credentials.getGrantType() == null) {
```
vikasrathee-cs, line 88: Will there be a field on UI for user to enter grant type.

vanathi-g: No, we are populating grant type based on the set of parameters being provided by the user

src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
vanathi-g: Removed it, thanks

src/main/java/io/cdap/plugin/salesforce/authenticator/Authenticator.java
vikasrathee-cs: wouldn't it be confusing for user to enter only partial details, it would be better from UI perspective if we can hide the fields which are not needed for a specific grant type.

vanathi-g: The Authenticator class fields are directly exposed to the UI? I thought that was only via the plugin configs

vikasrathee-cs: no, its not exposed, but in case user wants to use client_credentials grant, username password fields will still be visible

vanathi-g: I didn't add the field to all the plugin configs because I was unsure if the grant type itself is something that we expose to the user. But if it is required, we can change all the configs too to have this additional field.

vikasrathee-cs: its better to do it. @MrRahulSharma you can confirm. Salesforce has documented these grant types in detail.

MrRahulSharma: The plan currently is not to show this to CDF users, and only have the grant implementation for internal usage. As AFAIR there are no customer ask requiring this grant in CDF.

src/main/java/io/cdap/plugin/salesforce/authenticator/AuthenticatorCredentials.java
```diff
@@ -113,6 +110,62 @@ public String getProxyUrl() {
...
+
+  /**
+   * Create an instance of {@link AuthenticatorCredentials} from given set of parameters. Uses "password" grant type
+   * if username and password are non-null. Uses client-credentials grant type otherwise.
+   */
+  public static AuthenticatorCredentials fromParameters(String username, String password,
+                                                        String consumerKey, String consumerSecret, String loginUrl,
+                                                        Integer connectTimeout, Integer readTimeout, String proxyUrl) {
+    AuthenticatorCredentials.Builder builder;
+    if (username != null && password != null) {
```
janakigg, line 152: Should we add a unit test for this?

vikasrathee-cs:
@vanathi-g Add about new authentication mechanism in md files also.

src/main/java/io/cdap/plugin/salesforce/authenticator/AuthenticatorCredentials.java
vanathi-g: Done

MrRahulSharma (approved):

vanathi-g:
Looks like the e2e failures are unrelated to the code changes, merging this

PR #256, opened 2024-08-22
Added hook for treating all fields as nullable

MrRahulSharma:
Changing the code to make sure that all the fields are converted to Nullable based on the hook. In the previous implementation this was only added for Custom objects.


vikasrathee-cs (approved):
LGTM

albertshau (approved):

PR #266, opened 2024-10-15
Fix TaskAttempt Context is null

Vipinofficial11:
Changes include:

1. Fixes in the ```SalesforceWideRecordReader``` partitioning logic to address API retrieve limit exceeded errors.
2. Usage of default configurations in case of Non-CDAP calls

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
@@ -70,7 +70,10 @@ public class SalesforceWideRecordReader extends SalesforceBulkRecordReader {
   public SalesforceWideRecordReader initialize(
       InputSplit inputSplit, AuthenticatorCredentials credentials)
       throws IOException, InterruptedException {
-    List<Map<String, ?>> fetchedIdList = fetchBulkQueryIds(inputSplit, null);
+    // Use default configurations of BulkRecordReader.
+    super.initialize(inputSplit, credentials);
+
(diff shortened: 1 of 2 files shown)


MrRahulSharma:
Can we validate the WideRecordReader changes in CDAP/CDF today ?

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
```diff
@@ -70,7 +70,7 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
   public SalesforceWideRecordReader initialize(
       InputSplit inputSplit, AuthenticatorCredentials credentials)
       throws IOException, InterruptedException {
-    List<Map<String, ?>> fetchedIdList = fetchBulkQueryIds(inputSplit, null);
+    List<Map<String, ?>> fetchedIdList = fetchBulkQueryIds(inputSplit, credentials);
```
MrRahulSharma: Move the super.initialize call outside fetchBulkQueryIds and call it before the fetchBulkQueryIds call here.

Vipinofficial11: Moved ```super.initialize```  before ```fetchBulkQueryIds``` and removed paramters from ```fetchBulkQueryIds``` as they are not required anymore in the method. [b5c9fe5d483b43f874cdf4ade9370479ca92b1c0]

MrRahulSharma (approved):
LGTM

PR #275, opened 2025-04-29
Fixes addressing two issues identified during DTS transfer

Shubhangi-cs:
Fixes addressing two issues identified during transfer runs in the customer environment:

1. handling of `null` `XmlObject` instances that previously caused failures.
2. schema mismatches were between declared schema types and actual values resulted in casting exceptions for INT types.


harshdeeppruthi (approved):
Did we patch the customer and try out the fix ? If we have, we can merge this.

anup-cloudsufi:
> Did we patch the customer and try out the fix ? If we have, we can merge this.

Yes, these changes were patched and verified.

PR #287, opened 2025-06-10
[PLUGIN-1899] Added BulkConnection class with handling for connection leakage.

Krish-cloudsufi:
[PLUGIN-1899] Added Wrapper class for BulkConnection to handle connection leakage.

[PLUGIN-1899]: https://cdap.atlassian.net/browse/PLUGIN-1899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/com/sforce/async/BulkConnection.java
+++ b/src/main/java/com/sforce/async/BulkConnection.java
@@ -1,0 +1,844 @@
+/*
+ * Copyright © 2025 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * For implementation details, see {@link https://github.com/forcedotcom/wsc/blob/62.0.0/src/main/java/com/sforce/async/BulkConnection.java BulkConnection on GitHub}.
+ */
+package com.sforce.async;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.xml.namespace.QName;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.io.ByteStreams;
+import com.sforce.ws.ConnectionException;
+import com.sforce.ws.ConnectorConfig;
+import com.sforce.ws.MessageHandler;
+import com.sforce.ws.MessageHandlerWithHeaders;
+import com.sforce.ws.bind.CalendarCodec;
+import com.sforce.ws.bind.TypeMapper;
+import com.sforce.ws.parser.PullParserException;
+import com.sforce.ws.parser.XmlInputStream;
+import com.sforce.ws.parser.XmlOutputStream;
+import com.sforce.ws.transport.Transport;
+import com.sforce.ws.util.FileUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * BulkConnection
+ *
+ * @author mcheenath
+ * @since 160
+ */
+public class BulkConnection {
+    public static final String NAMESPACE = "http://www.force.com/2009/06/asyncapi/dataload";
+    public static final String SESSION_ID = "X-SFDC-Session";
+    public static final String XML_CONTENT_TYPE = "application/xml";
+    public static final String CSV_CONTENT_TYPE = "text/csv";
+    public static final String JSON_CONTENT_TYPE = "application/json";
+    public static final String ZIP_XML_CONTENT_TYPE = "zip/xml";
+    public static final String ZIP_CSV_CONTENT_TYPE = "zip/csv";
+    public static final String ZIP_JSON_CONTENT_TYPE = "zip/json";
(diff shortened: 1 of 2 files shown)


src/main/java/com/sforce/async/BulkConnection.java
harshdeeppruthi, line 1: I see a lot of formatting changes in this file (probably due to AutoFormat ?)
Since this is not directly from our codebase, we should make only minimal changes.
My suggestions are:
- Clearly point out the version we're copying from the original repo. I'm guessing it is for [v62](https://github.com/forcedotcom/wsc/blob/62.0.0/src/main/java/com/sforce/async/BulkConnection.java) ?
- Avoid any other formatting changes. The only diff from the original file should be the newly added lines.
- Clearly point out what lines have been edited/added using a comment suffix.

harshdeeppruthi (changes requested):

src/main/java/com/sforce/async/BulkConnection.java
Shubhangi-cs: Yes the file is same as of connectors. The formatting is being changes as per the checkStyle. 
Clearly point out the version we're copying from the original repo. I'm guessing it is for [v62](https://github.com/forcedotcom/wsc/blob/62.0.0/src/main/java/com/sforce/async/BulkConnection.java) ?
Please elaborate more on this, if this is something related to the version upgrade then it is still 53 as we have reverted it due to backward compatibility

Shubhangi-cs: @harshdeeppruthi changed the BulkConnection class as per V62 and also tried to keep formatting changes to minimal

pom.xml
harshdeeppruthi, line 1: We don't need this change anymore I think ?

src/main/java/com/sforce/async/BulkConnection.java
```diff
@@ -0,0 +1,844 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Every line changed/added is suffixed with a comment // %EDIT%
+ * Author: @harshpruthi
+ */
+package com.sforce.async;
+import com.fasterxml.jackson.core.JsonFactory;
```
harshdeeppruthi: The import order of these seem to have moved up w.r.t to the [original file](https://github.com/forcedotcom/wsc/blob/62.0.0/src/main/java/com/sforce/async/BulkConnection.java). Please keep it as it is.

src/main/java/com/sforce/async/BulkConnection.java
```diff
@@ -0,0 +1,844 @@
+/*
```
harshdeeppruthi, line 1: Not sure myself, but wondering if we're allowed to change the license like this. Can we confirm if this is allowed

src/main/java/com/sforce/async/BulkConnection.java
```diff
@@ -0,0 +1,844 @@
...
+        } catch (PullParserException e) {
+            throw new AsyncApiException("Failed to parse batch info ", AsyncExceptionCode.ClientInputError, e);
+        } catch (ConnectionException e) {
+            throw new AsyncApiException("Failed to parse batch info ", AsyncExceptionCode.ClientInputError, e);
+        } finally {                                                                             // %EDIT%
+            if (stream != null) {                                                                 // %EDIT%
+                try {                                                                               // %EDIT%
+                    stream.close();                                                                   // %EDIT%
+                } catch (IOException e) {                                                           // %EDIT%
+                    logger.warn("Failed to close input stream with error: {}, cause: {}", e.getCause(), e.getCause());// %EDIT%
```
harshdeeppruthi: nit: Here and probably at a few other places, we're logging cause twice. Please fix this.

src/main/java/com/sforce/async/BulkConnection.java
itsankit-google: Yeah this is allowed.

src/main/java/com/sforce/async/BulkConnection.java
itsankit-google: The plugin files need to follow google configuration checker style. We can add a javadoc that the file is copied from library with a link.

pom.xml
Shubhangi-cs: removed

src/main/java/com/sforce/async/BulkConnection.java
Shubhangi-cs: updated

src/main/java/com/sforce/async/BulkConnection.java
Shubhangi-cs: updated

src/main/java/com/sforce/async/BulkConnection.java
harshdeeppruthi: Specifically regarding @itsankit-google's comment, I am of the opinion we should keep our own formatting changes to the minimum possible (and suppress any formatting checks on this file) since this is not our source and it will be easier to maintain in the future if we keep a minimum required diff. What do you think ?

Shubhangi-cs: Updated file has minimum changes as all the formatting checks is being suppressed for this file.

src/main/java/com/sforce/async/BulkConnection.java
```diff
@@ -0,0 +1,845 @@
...
+        } catch (PullParserException e) {
+            throw new AsyncApiException("Failed to parse batch info ", AsyncExceptionCode.ClientInputError, e);
+        } catch (ConnectionException e) {
+            throw new AsyncApiException("Failed to parse batch info ", AsyncExceptionCode.ClientInputError, e);
+        } finally {                                                                             // %EDIT%
+            if (stream != null) {                                                                 // %EDIT%
+                try {                                                                               // %EDIT%
+                    stream.close();                                                                   // %EDIT%
+                } catch (IOException e) {                                                           // %EDIT%
+                    logger.warn("Failed to close input stream with error: {}, cause: {}", e.getCause());// %EDIT%
```
harshdeeppruthi: I think we should be logging both message and cause ?

src/main/java/com/sforce/async/BulkConnection.java
```diff
@@ -0,0 +1,845 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Every line changed/added is suffixed with a comment // %EDIT%
+ * Author: @harshpruthi
```
harshdeeppruthi: nit: We can remove this. Also, let's add a link to the original file: https://github.com/forcedotcom/wsc/blob/62.0.0/src/main/java/com/sforce/async/BulkConnection.java in a comment

harshdeeppruthi (approved):
Left a couple nit comments, please address them before merging.
Also, I see e2e tests are failing. Please take a look into that before merging as well.

src/main/java/com/sforce/async/BulkConnection.java
Shubhangi-cs: done

src/main/java/com/sforce/async/BulkConnection.java
Shubhangi-cs: done

Shubhangi-cs:
> Left a couple nit comments, please address them before merging. Also, I see e2e tests are failing. Please take a look into that before merging as well.

Addressed all the comments

PR #288, opened 2025-06-10
Wide Record reader fix.

prince-cs:
Currently records from a single batch are being read in a parallel stream processing and the result is being stored in its entirety. This could lead to out of memory(OOM). 

The fix is to ensure that the batches are being read with the help of an iterator rather storing the entire result. Made the changes accordingly to `SalesforceWideRecordReader.java` 

Validation has been done by making a query of length more than 20k and ensuring the pipeline passes.

JIRA: https://cdap.atlassian.net/browse/PLUGIN-1897

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceConstants.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceConstants.java
@@ -47,7 +47,7 @@ public class SalesforceConstants {
   public static final int RANGE_FILTER_MIN_VALUE = 0;
   public static final int SOQL_MAX_LENGTH = 20000;
 
-  public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 30000;
+  public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 120_000;
   public static final int DEFAULT_READ_TIMEOUT_SEC = 18000;
   public static final String PROPERTY_CONNECT_TIMEOUT = "connectTimeout";
   public static final String PROPERTY_READ_TIMEOUT = "readTimeout";
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
@@ -34,6 +34,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -48,10 +49,17 @@ public class SalesforceWideRecordReader extends SalesforceBulkRecordReader {
 
   private final String query;
   private final SoapRecordToMapTransformer transformer;
+  private Iterator<List<Map<String, ?>>> batchIterator;
 
   private List<Map<String, ?>> results;
   private Map<String, ?> value;
   private int index;
+  private AuthenticatorCredentials credentials;
+  private PartnerConnection partnerConnection;
+  private SObjectDescriptor sObjectDescriptor;
+  private List<String> fieldsNames;
+  private String fields;
+  private String sObjectName;
 
   public SalesforceWideRecordReader(Schema schema, String query, SoapRecordToMapTransformer transformer) {
     super(schema);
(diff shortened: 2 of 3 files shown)


MrRahulSharma:
Please add more details in the description, like the issue, RCA, fix, how did we validate etc.

src/main/java/io/cdap/plugin/salesforce/SalesforceConstants.java
```diff
@@ -47,7 +47,7 @@ public class SalesforceConstants {
   public static final int RANGE_FILTER_MIN_VALUE = 0;
   public static final int SOQL_MAX_LENGTH = 20000;
 
-  public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 30000;
+  public static final int DEFAULT_CONNECTION_TIMEOUT_MS = 120_000;
```
MrRahulSharma, line 50: This change does not look related to this failure, can you confirm ?

anup-codex: The read calls started failing with `timeouts` after reading some batches and did not recover even after retries.
During local testing faced the same behavior in both CDF and DTS hence increasing the timeout helped overcome the timeouts. 
This was observed in our environment.

itsankit-google (commented):
please add JIRA in the PR title.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
```diff
@@ -48,10 +49,17 @@ public class SalesforceWideRecordReader extends SalesforceBulkRecordReader {
...
 
   private List<Map<String, ?>> results;
   private Map<String, ?> value;
   private int index;
+  private AuthenticatorCredentials credentials;
+  private PartnerConnection partnerConnection;
+  private SObjectDescriptor sObjectDescriptor;
+  private List<String> fieldsNames;
+  private String fields;
+  private String sObjectName;
```
itsankit-google, line 62: why all of them need to be added as instance variables?

prince-cs: as we are now fetching records based on batch, so it is required to declare these as instance variables

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
```diff
@@ -73,33 +81,21 @@ public SalesforceWideRecordReader initialize(
...
-      List<List<Map<String, ?>>> partitions =
-        Lists.partition(fetchedIdList, SalesforceSourceConstants.WIDE_QUERY_MAX_BATCH_COUNT);
+      partnerConnection = SalesforceConnectionUtil.getPartnerConnection(credentials);
+        sObjectDescriptor = SObjectDescriptor.fromQuery(query);
+      fieldsNames = sObjectDescriptor.getFieldsNames();
+      fields = String.join(",", fieldsNames);
+      sObjectName = sObjectDescriptor.getName();
       LOG.debug("Number of partitions to be fetched for wide object: '{}'", partitions.size());
-
-      // Process partitions with batches sized to adhere to API limits and optimize memory usage.
```
itsankit-google, line 90: can we hit API limit early now after this change?

anup-codex: No, as the logic remains the same, the only change is the way it reads the set of 2K records at a time instead of the entire batch in one go.

src/main/java/io/cdap/plugin/salesforce/SalesforceConstants.java
MrRahulSharma: We do have a user provided value for the connectTimeout, are we not using that for some flow ?

anup-codex: Yes, the plugin does use the user provided `connectTimeout`. However it fails with the default `connectTimeout`, so increased the timeout to handle the default configs.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceWideRecordReader.java
MrRahulSharma, line 1: There are other places in this file which can also be optimized, such as in the processPartition method why do we need to call fetchPartition API for every sObjectId ? Ideally since this API expects a list of the sObjectIds, hence we should be making a single call to fetch multiple sObjects.

prince-cs: As discussed, I ran the pipeline in debug mode to check the values being passed and it looked like that fetchPartition was being called every SObjectId but in actuality it is being populated with an array of SObjectIds. So, the logic is fine.

MrRahulSharma (commented):
LGTM

MrRahulSharma (approved):

PR #289, opened 2025-06-11
"apiVersion with backward compatibility"

Shubhangi-cs:
Make changes in Salesforce plugin to read from either v53 or v62 based on provided value, making apiversion parameterised

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <name>Salesforce plugins</name>
   <groupId>io.cdap.plugin</groupId>
   <artifactId>salesforce-plugins</artifactId>
-  <version>1.6.11</version>
+  <version>1.6.12-SNAPSHOT</version>
   <packaging>jar</packaging>
   <description>Salesforce Plugins</description>
   <url>https://github.com/data-integrations/salesforce</url>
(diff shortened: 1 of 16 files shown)


src/main/java/io/cdap/plugin/salesforce/SalesforceConnectionUtil.java
```diff
@@ -41,7 +41,7 @@ public class SalesforceConnectionUtil {
    */
   public static PartnerConnection getPartnerConnection(AuthenticatorCredentials credentials)
```
vikasrathee-cs, line 42: make sure we are not calling this anywhere in record reader

Shubhangi-cs: Being used in soapRecReader and WideRecReader

pom.xml
```diff
@@ -22,7 +22,7 @@
   <name>Salesforce plugins</name>
   <groupId>io.cdap.plugin</groupId>
   <artifactId>salesforce-plugins</artifactId>
-  <version>1.6.11</version>
+  <version>1.6.11-SNAPSHOT</version>
```
MrRahulSharma: This should be 1.6.12-SNAPSHOT

MrRahulSharma:
I am not sure that I understand the intent of the change. Based on my understanding, what we really want to achieve is that the given plugin codebase code should work using a specific version number set for the plugin, and default to a version in case the no version number is set for the plugin. 
I see that in some APIs we are using SalesforceConstants.API_VERSION and in some we are using the apiVersion parameter in the API signature. Which can lead to weird issues in future in which different calls get invoked on different api versions. Hence, can we have a consistent mechanism put in place which ensures that all the salesforce api calls will consistently use a single version number, which is either provided or the default.

pom.xml
Shubhangi-cs: updated

vikasrathee-cs:
closing this PR as this should be handled on platform side.

PR #312, opened 2025-07-17
[PLUGIN-1904] Fix: Vulnerability for json_version

AbhishekKumar9984:
Issue:
Addressed Denial of Service (DoS) vulnerabilities in the org.json:json library and related components. JSON-Java versions up to and including 20230618 contained a parser bug allowing modestly sized input to cause excessive memory consumption. Additionally, a stack overflow in the XML.toJSONObject component of hutool-json v5.8.10 and org.json:json versions prior to 20231013 allowed attackers to exploit crafted JSON or XML data to trigger DoS.

Root Cause :
Insufficient validation and resource management in the parser enabled malicious input to exhaust system memory or cause stack overflows.

Fix : Upgraded org.json:json to a secure version (20231013) .

JIRA Ticket : https://cdap.atlassian.net/browse/PLUGIN-1904

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceUtil.java
@@ -33,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import scala.reflect.ClassTag$;
 
+import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
@@ -146,6 +147,13 @@ final class SalesforceStreamingSourceUtil {
           String.format("Field '%s' is of type '%s', but value found is '%s'",
                         field.getName(), fieldSchemaType, value));
       }
+    }
+
+    // NOTE: org.json >= 20230227 returns BigDecimal for all non-integer JSON numbers.
+    if (value instanceof BigDecimal && fieldSchemaType.equals(Schema.Type.DOUBLE)) {
+      // Avro Schema.Type.DOUBLE expects a Double instance (or primitive double) at serialization time,
+      // so converting BigDecimal → double for compatibility.
+      return ((BigDecimal) value).doubleValue();
     }
 
     return value;
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <commons.csv.version>1.6</commons.csv.version>
     <jackson.version>1.9.13</jackson.version>
     <jackson2.version>2.17.1</jackson2.version>
-    <json.version>20180813</json.version>
+    <json.version>20231013</json.version>
     <awaitility.version>3.1.6</awaitility.version>
     <commons-logging.version>1.2</commons-logging.version>
     <testSourceLocation>${project.basedir}/src/test/java/</testSourceLocation>


src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceUtil.java
```diff
@@ -148,6 +149,13 @@ private static Object convertValue(Object value, Schema.Field field) {
       }
     }
 
+    // NOTE: org.json >= 20230227 returns BigDecimal for all non-integer JSON numbers.
```
vikasrathee-cs, line 152: Have we checked the backward compatibility for this?

vikasrathee-cs (approved):

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceUtil.java
AbhishekKumar9984: yes , the fix is backward compatible.

sgarg-CS (approved):
LGTM

src/main/java/io/cdap/plugin/salesforce/plugin/source/streaming/SalesforceStreamingSourceUtil.java
minurajeeve: What was the behaviour previously ie in 20180813 -> were we getting double for all non-integer JSON numbers?

Also can you please explain how you checked backward compatibility for this? 

Some concerns I have (AI generated)
1. Loss of Precision
A BigDecimal can represent decimal numbers with arbitrary precision, while a double has a fixed precision (about 15-17 decimal digits). If the BigDecimal holds a value with more precision than a double can handle, the extra information will be lost (rounded) during the conversion.

Example: Consider a BigDecimal representing a highly precise scientific measurement.

```
BigDecimal preciseValue = new BigDecimal("0.1234567890123456789");
double doubleValue = preciseValue.doubleValue();
// doubleValue is now 0.12345678901234568
```
The original precision is lost, which can be critical in financial or scientific applications where exact values are paramount.

2. Range and Overflow Issues
A BigDecimal can represent a much larger range of numbers than a double. If the BigDecimal's value exceeds the maximum or minimum value that a double can store, the conversion will result in an overflow or underflow.

Example: A number larger than Double.MAX_VALUE (~1.8 x 10^{308}).

```
// A number that is too large for a double
BigDecimal hugeValue = new BigDecimal("2.0e308");
double doubleValue = hugeValue.doubleValue();
// doubleValue becomes Double.POSITIVE_INFINITY
```
In this case, the actual numeric value is replaced with a special Infinity value, which is a significant loss of information. Similarly, a BigDecimal value smaller in magnitude than Double.MIN_NORMAL can lose precision and become zero.

Have we tested for these edge cases?

AbhishekKumar9984: yes , the behaviour previously ie in 20180813 ->  we were getting double for all non-integer JSON numbers.

We are converting `BigDecimal` to `Double` to maintain compatibility with existing Avro `"double"` schema behavior prior to the `org.json` upgrade. This preserves the historical behavior (including precision limits of double) since `org.json` previously produced `Double` directly.

There is no schema change here, we're just making the value type align with what the schema already expects.

**Before the org.json upgrade**
- `org.json` was giving us a Double schema already.
- If the original JSON contained a number such as `"0.1234567890123456789"`, `org.json` already rounded it when creating the Double internally, so precision loss was already happening.
- Same thing for large values: overflow to Infinity or underflow to 0 was part of the earlier version.

**After the org.json upgrade**
- Now `org.json` gives you a BigDecimal instead.
- If we convert it to Double at serialization time, we’re basically reproducing the same precision and range limitations that existed from before.

Therefore, for existing customers who were already consuming Avro double fields, the observable behavior remains the same as it was before the upgrade.


[BEFORE]
image


[AFTER]
image


Also , Salesforce itself supports a maximum of 17 digits after the decimal — anything more gets rounded off at the data source level.

image

minurajeeve (approved):

PR #317, opened 2025-07-25
Fix: Vulnerability for guava

Krish-cloudsufi:
Issue:
Resolved a high-severity information disclosure vulnerability (CVE-2023-2976) related to insecure temporary file handling in the FileBackedOutputStream class of the Guava library (com.google.guava:guava). This vulnerability affects versions 1.0 through 31.1 on Unix-based systems and Android Ice Cream Sandwich. In vulnerable versions, temporary files were created in the system's default temporary directory (/tmp), allowing local users or applications with access to the directory to potentially read or tamper with those files.

Root Cause:
Guava used Java's default temporary directory for storing files without proper isolation, which could expose sensitive data in shared environments.

Fix:
Removed the explicit dependency on Guava from the Salesforce plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -754,14 +754,6 @@
         </plugins>
       </build>
       <dependencyManagement>
-        <dependencies>
-          <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>31.0.1-jre</version>
-            <scope>compile</scope>
-          </dependency>
-        </dependencies>
       </dependencyManagement>
       <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,16 @@
     </repository>
   </repositories>
 
+  <dependencyManagement>
+    <dependencies>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -754,14 +754,6 @@
...
       </build>
       <dependencyManagement>
-        <dependencies>
-          <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>31.0.1-jre</version>
-            <scope>compile</scope>
-          </dependency>
-        </dependencies>
```
sgarg-CS: Can we skip this as it is not showing up in the list of guava CVE's
`/opt/connectors/lib/salesforce/libs/io.cdap.plugin.salesforce-plugins-1.6.13.jar/lib/guava-13.0.1.jar`

image

sgarg-CS: Also, exclude the transitive dependency fetched through `hydrator-common`:`2.10.0` and explicitly add the upgraded version.
image

PR #326, opened 2025-08-06
[PLUGIN-1891] Updated the PK chunking supported object list.

vikasrathee-cs:
Updated the PK chunking supported object list. This change will go with 53 to 62 upgrade.


MrRahulSharma (approved):
LGTM

PR #336, opened 2025-09-29
Moved the pk chunking check to SplitUtil class

vikasrathee-cs:
Moved the pk chunking check to SplitUtil class

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.salesforce.plugin.source.batch.util;
 
+import com.google.common.base.Strings;
 import com.sforce.async.AsyncApiException;
 import com.sforce.async.AsyncExceptionCode;
 import com.sforce.async.BatchInfo;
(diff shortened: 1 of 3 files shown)


src/test/java/io/cdap/plugin/salesforce/SalesforceBulkUtilTest.java
```diff
@@ -83,4 +85,17 @@ public void testAwaitCompletionBatchCompleted() throws Exception {
     // Call the awaitCompletion method and verify that it completes successfully
     SalesforceBulkUtil.awaitCompletion(bulkConnection, job, Collections.singletonList(batchInfo), true);
   }
+
+  @Test
```
MrRahulSharma, line 90: Add a UT to validate that there are no entries with the upper case values.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -252,4 +252,16 @@ public static RetryPolicy<Object> getRetryPolicy(Long initialRetryDuration, Long
       return RetryPolicy.builder().withMaxRetries(0).build();
     }
   }
+
+  // This is added for UCS use case only to identify the objects where PK chunking needs to be enabled by default.
+  public static boolean isPkChunkingSupported(String sobjectName) {
+    return sobjectName != null && (
```
MrRahulSharma: Add a check for empty sObjectName

MrRahulSharma: Better break this check into multiple cases for better readability.

src/test/java/io/cdap/plugin/salesforce/SalesforceBulkUtilTest.java
vikasrathee-cs: Added

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
vikasrathee-cs: Done

MrRahulSharma (approved):

PR #353, opened 2026-03-19
[PLUGIN-1953] Support for client credentials Grant type

vikasrathee-cs:
# Support for client credentials Grant type

Jira : [PLUGIN-1953](https://cdap.atlassian.net/browse/PLUGIN-1953)

## Summary

Adds support for the **Client Credentials** OAuth grant type to all Salesforce plugins (source, sink, streaming, connector).

### Widgets (UI)
Expose the **OAuth Grant Type** radio group and add a filter that hides `username` / `password` / `securityToken` when `client_credentials` is selected.
- `widgets/Salesforce-connector.json`
- `widgets/Salesforce-batchsink.json`
- `widgets/Salesforce-batchsource.json`
- `widgets/Salesforce-streamingsource.json`
- `widgets/SalesforceMultiObjects-batchsource.json`

### Docs
Document the new **Grant Type** field, clarify which fields are required per grant type, and note that `client_credentials` requires an instance-specific login URL.
- `docs/Salesforce-connector.md`
- `docs/Salesforce-batchsink.md`
- `docs/Salesforce-batchsource.md`
- `docs/Salesforce-streamingsource.md`
- `docs/SalesforceMultiObjects-batchsource.md`

## What's new

**New config property `authenticationGrantType`** backed by the pre-existing `AuthenticatorCredentials.GrantType` enum (`password` / `client_credentials`). Null is treated as `password` so upgraded pipelines that were saved before this change keep working without manual edits.

### Validation
New `validateAuthenticationFields(FailureCollector)` on `SalesforceConnectorBaseConfig`:
- Always required: `consumerKey`, `consumerSecret`, `loginUrl`.
- Only required for `password`: `username`, `password`, `securityToken`.
- Macro-aware skips any field whose value is a macro, and skips entirely if the grant type itself is a macro.
- Each missing field reports its own failure tied to its `withConfigProperty(...)`, so the UI highlights the exact field.

### UI changes
- New **OAuth Grant Type** radio group (inline layout, defaults to `password`) added to all five widget JSONs.
- A `GrantTypePassword` filter hides `username` / `password` / `securityToken` when `client_credentials` is selected, so users only see the fields they need to fill in.

### Screenshots

- Radio [Password]
image

- Radio [Client Credentials]
image

- Missing value validation
image

## Tests added

New file: **`SalesforceConnectorBaseConfigTest.java`** (250 lines) covering `validateAuthenticationFields()`:

- **Password grant** happy path + one test per missing field (`consumerKey`, `consumerSecret`, `loginUrl`, `username`, `password`, `securityToken`), including empty-string handling, plus an all-missing test that asserts exactly 6 failures on the right fields.
- **Client Credentials grant** happy path, missing `consumerKey` / `consumerSecret` / `loginUrl`, explicit assertion that `username` / `password` / `securityToken` are *not* flagged, and an all-missing test that asserts exactly 3 failures.
- **Default behavior**  null `authenticationGrantType` falls back to `PASSWORD` validation rules.

## Additional notes for reviewer

- **Backward compatibility:** existing pipelines persisted without `authenticationGrantType` will deserialize with a null value, which `getAuthenticationGrantType()` maps to `PASSWORD`.

- **Login URL gotcha:** `client_credentials` requires an instance-specific URL (e.g. `https://<instance>.my.salesforce.com/services/oauth2/token`) rather than the generic `login.salesforce.com` endpoint. This is called out in the updated docs.

[PLUGIN-1953]: https://cdap.atlassian.net/browse/PLUGIN-1953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.salesforce.plugin;
 
+import com.google.common.base.Strings;
 import com.sforce.ws.ConnectionException;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
@@ -24,6 +25,7 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
 import io.cdap.plugin.salesforce.SalesforceConstants;
 import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
+import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials.GrantType;
 
 import javax.annotation.Nullable;
 
@@ -31,6 +33,12 @@ import javax.annotation.Nullable;
  * Base configuration for Salesforce Streaming and Batch plugins
  */
 public class SalesforceConnectorBaseConfig extends PluginConfig {
+
+  @Name(SalesforceConstants.PROPERTY_AUTHENTICATION_GRANT_TYPE)
+  @Description("Salesforce authentication grant type: basic or client credentials")
+  @Nullable
+  @Macro
+  protected String authenticationGrantType;
 
   @Nullable
   @Name(SalesforceConstants.PROPERTY_PROXY_URL)
@@ -81,7 +89,11 @@ public class SalesforceConnectorBaseConfig extends PluginConfig {
   private final String securityToken;
 
   @Name(SalesforceConstants.PROPERTY_LOGIN_URL)
-  @Description("Endpoint to authenticate to")
+  @Description("Salesforce OAuth2 login URL. For the 'password' grant type, the default generic URL\n" +
+          "`https://login.salesforce.com/services/oauth2/token` can be used. " +
+          "For the 'client_credentials' grant type,\n" +
+          "you must provide your Salesforce instance-specific URL, for example\n" +
+          "`https://<your-instance>.my.salesforce.com/services/oauth2/token`.")
   @Macro
   @Nullable
   private final String loginUrl;
@@ -118,7 +130,8 @@ public class SalesforceConnectorBaseConfig extends PluginConfig {
                                        @Nullable Long initialRetryDuration,
                                        @Nullable Long maxRetryDuration,
                                        @Nullable Integer maxRetryCount,
-                                       @Nullable Boolean retryOnBackendError) {
+                                       @Nullable Boolean retryOnBackendError,
+                                       @Nullable String authenticationGrantType) {
     this.consumerKey = consumerKey;
     this.consumerSecret = consumerSecret;
     this.username = username;
@@ -132,6 +145,16 @@ public class SalesforceConnectorBaseConfig extends PluginConfig {
     this.maxRetryDuration = maxRetryDuration;
     this.retryOnBackendError = retryOnBackendError;
     this.maxRetryCount = maxRetryCount;
+    this.authenticationGrantType = authenticationGrantType;
+  }
+
+  public GrantType getAuthenticationGrantType() {
+    if (!Strings.isNullOrEmpty(authenticationGrantType) &&
+            authenticationGrantType.equals(GrantType.CLIENT_CREDENTIALS.getType())) {
+      return GrantType.CLIENT_CREDENTIALS;
+    }
+    // Default auth, handles null case when upgrading pipeline
+    return SalesforceConstants.DEFAULT_GRANT_TYPE;
   }
 
   @Nullable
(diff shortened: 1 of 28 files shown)


itsankit-google (commented):
it is okay to remove null check because it is not required in certain cases.

But we should also add validation for the type when these configs are required.

vikasrathee-cs:
> it is okay to remove null check because it is not required in certain cases.
> 
> But we should also add validation for the type when these configs are required.

As of now, there is no explicit variable defined which calculates whether user is supporting password grant or client_creds. Based on the values, if user name and password are null, it will fallback to client creds automatically. This was done to support DTS use case I guess.

itsankit-google:
> > it is okay to remove null check because it is not required in certain cases.
> > But we should also add validation for the type when these configs are required.
> 
> As of now, there is no explicit variable defined which calculates whether user is supporting password grant or client_creds. Based on the values, if user name and password are null, it will fallback to client creds automatically. This was done to support DTS use case I guess.

As discussed offline, this way we will not be able to maintain the extensibility of CDF plugin's ability to support multiple auth types.

We can add a `AuthenticationType` similar to [google-cloud](https://github.com/data-integrations/google-cloud/blob/5bb35a4b7aba252f4c16dec7106a710d45878a7f/widgets/GCS-batchsink.json#L45) plugins to correctly fix this.

We can also hide certain fields in CDF Plugin UI based on `authenticationType`.

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceOutputFormatProvider.java
```diff
@@ -75,13 +75,22 @@ public SalesforceOutputFormatProvider(SalesforceSinkConfig config) {
...
     } else {
-      configBuilder
-        .put(SalesforceConstants.CONFIG_USERNAME, Objects.requireNonNull(config.getConnection().getUsername()))
-        .put(SalesforceConstants.CONFIG_PASSWORD, Objects.requireNonNull(config.getConnection().getPassword()))
-        .put(SalesforceConstants.CONFIG_CONSUMER_KEY, Objects.requireNonNull(config.getConnection().getConsumerKey()))
-        .put(SalesforceConstants.CONFIG_CONSUMER_SECRET, Objects.requireNonNull(config.getConnection().
-                                                                                  getConsumerSecret()))
-        .put(SalesforceConstants.CONFIG_LOGIN_URL, Objects.requireNonNull(config.getConnection().getLoginUrl()));
+      switch (config.getConnection().getAuthenticationGrantType()) {
+        case BASIC_AUTHENTICATION:
```
vikasrathee-cs: there is no basic auth in salesforce, previously password grant type was supported and now it is client creds also

src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
```diff
@@ -132,6 +141,16 @@ public SalesforceConnectorBaseConfig(@Nullable String consumerKey,
     this.maxRetryDuration = maxRetryDuration;
     this.retryOnBackendError = retryOnBackendError;
     this.maxRetryCount = maxRetryCount;
+    this.authenticationGrantType = authenticationGrantType;
+  }
+
+  public AuthenticationGrantType getAuthenticationGrantType() {
+    if (!Strings.isNullOrEmpty(authenticationGrantType) &&
+        authenticationGrantType.equals(AuthenticationGrantType.BASIC_AUTHENTICATION.getValue())) {
```
vikasrathee-cs: make it password grant type as default

widgets/Salesforce-batchsink.json
```diff
@@ -266,6 +285,42 @@
     }
   ],
   "filters": [
+    {
+      "name": "GrantTypeBasicAuthentication",
+      "condition": {
+        "expression": "authenticationGrantType == 'basicAuthentication'"
```
vikasrathee-cs: make it not clientCredentialsAuthentication, then it will be backward compatible.

vikasrathee-cs (commented):
change the auth type from based to password grant type

src/main/java/io/cdap/plugin/salesforce/AuthenticationGrantType.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright © 2026 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
```
vikasrathee-cs: AuthenticatorCredentials class already has a GrantType available which is getting used in DTS also, it should be set based on this UI property.

That existing class can be used instead of this.

src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
psainics: this is handeled by the else case

```java
// Default auth, handles null case when upgrading pipeline
    return SalesforceConstants.DEFAULT_GRANT_TYPE
```

widgets/Salesforce-batchsink.json
psainics: updated!

src/main/java/io/cdap/plugin/salesforce/AuthenticationGrantType.java
psainics: removed this class

src/main/java/io/cdap/plugin/salesforce/plugin/sink/batch/SalesforceOutputFormatProvider.java
```diff
@@ -75,13 +76,15 @@ public SalesforceOutputFormatProvider(SalesforceSinkConfig config) {
...
-        .put(SalesforceConstants.CONFIG_CONSUMER_KEY, Objects.requireNonNull(config.getConnection().getConsumerKey()))
-        .put(SalesforceConstants.CONFIG_CONSUMER_SECRET, Objects.requireNonNull(config.getConnection().
-                                                                                  getConsumerSecret()))
-        .put(SalesforceConstants.CONFIG_LOGIN_URL, Objects.requireNonNull(config.getConnection().getLoginUrl()));
+      configBuilder.put(SalesforceConstants.CONFIG_CONSUMER_KEY,
+                      Objects.requireNonNull(config.getConnection().getConsumerKey()))
+              .put(SalesforceConstants.CONFIG_CONSUMER_SECRET,
+                      Objects.requireNonNull(config.getConnection().getConsumerSecret()))
+              .put(SalesforceConstants.CONFIG_LOGIN_URL, Objects.requireNonNull(config.getConnection().getLoginUrl()));
+      if (config.getConnection().getAuthenticationGrantType() == AuthenticatorCredentials.GrantType.PASSWORD) {
```
vikasrathee-cs, line 84: if grant type is password then only this configuration is mandatory

src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
```diff
@@ -232,4 +251,55 @@ public String getProxyUrl() {
...
+   * For CLIENT_CREDENTIALS grant type: consumerKey, consumerSecret, and loginUrl are required.
+   *
+   * @param collector the failure collector to report validation errors
+   */
+  public void validateAuthenticationFields(FailureCollector collector) {
+    if (containsMacro(SalesforceConstants.PROPERTY_AUTHENTICATION_GRANT_TYPE)) {
+      return;
+    }
+
+    GrantType grantType = getAuthenticationGrantType();
```
itsankit-google: move this to before line no 286 where the `grantType` is actually used.

src/main/java/io/cdap/plugin/salesforce/plugin/SalesforceConnectorBaseConfig.java
itsankit-google: please add more details about login Url based on auth type here as well

itsankit-google (approved):
couple of minor comments, rest LGTM.

Please squash commits before merge.

PR #354, opened 2026-04-20
[PLUGIN-1957] Validate PK Chunking for incremental loads

harishhk107:
# [PLUGIN-1957] Autodetect PK Chunking for incremental loads

## What

Adds an autonomous pre-flight check to decide whether to enable PK Chunking in `SalesforceBatchSource.getSplits()`.
*   If the estimated record count is below the user-configured chunk size (`config.getChunkSize()`) and the query is selective, PK chunking is skipped and executed as a standard un-chunked query to avoid "empty chunk" overhead on small datasets.
*   If the dataset volume exceeds the threshold, or is flagged as non-selective, PK Chunking is enabled.
*   This feature is specifically designed for automated Data Transfer Service.

## Why

PK chunking is designed for very large datasets. Enabling it on small datasets (like highly restrictive incremental intervals where 0 or very few records changed) causes dozens of empty chunk boundaries to be spooled and executed, wasting daily Bulk API allocations and adding massive pipeline overhead. This change ensures chunking is only applied when operationally justified by the actual matching record volume.

## Changes

*   **`SalesforceBatchSource.java`**: Configured PK chunking threshold dynamically using `chunkSize`.
*   **`SalesforceSplitUtil.java`**: Added REST Query Plan explain check and removed unused constants.
*   **`SalesforceQueryUtil.java`**: Replaced magic HTTP status codes with constants.
*   **`SalesforceSourceConstants.java`**: Cleaned up the unused threshold constant.
*   **`SalesforceQueryUtilTest.java`**: Added unit tests covering Query Plan and fallback paths.
*   **`SalesforceBatchSourceETLTest.java`**: Renamed ETL test case to remove UI naming.

---

[PLUGIN-1957]: https://cdap.atlassian.net/browse/PLUGIN-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
@@ -15,8 +15,22 @@
  */
 package io.cdap.plugin.salesforce;
 
+import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import io.cdap.plugin.salesforce.authenticator.Authenticator;
+import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
 import io.cdap.plugin.salesforce.parser.SalesforceQueryParser;
+import io.cdap.plugin.salesforce.plugin.OAuthInfo;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpMethod;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
 
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URLEncoder;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
 
@@ -84,6 +98,18 @@ public class SalesforceQueryUtil {
   }
 
   /**
+   * Creates a COUNT query from an existing SOQL query.
+   * Replaces the SELECT fields with COUNT() while preserving the FROM and WHERE clauses.
+   *
+   * @param query the original SOQL query
+   * @return a COUNT SOQL query string
+   */
+  public static String createCountQuery(String query) {
+    String fromStatement = SalesforceQueryParser.getFromStatement(query);
+    return SELECT + "COUNT() " + fromStatement;
+  }
+
+  /**
    * Generates SObject query filter based on provided values.
    *
    * @param filterDescriptor filter options holder
@@ -105,5 +131,86 @@ public class SalesforceQueryUtil {
         .append(filterDescriptor.getEndTime().format(DateTimeFormatter.ISO_DATE_TIME));
     }
     return filter.toString();
+  }
+
+  /**
+   * Fetches query plan from Salesforce query explain REST API.
+   *
+   * @param query SOQL query
+   * @param credentials credentials to authenticate the call
+   * @return query plan details
+   * @throws Exception in case of network or API failures
+   */
+  public static QueryPlanResponse getQueryPlan(
+      final String query,
+      final AuthenticatorCredentials credentials)
+      throws Exception {
+    OAuthInfo oAuthInfo = Authenticator.getOAuthInfo(credentials);
+    String explainUrl = String.format(
+        "%s/services/data/v%s/query/?explain=%s",
+        oAuthInfo.getInstanceURL(),
+        SalesforceConstants.API_VERSION,
+        URLEncoder.encode(query, "UTF-8"));
+
+    SslContextFactory sslContextFactory = new SslContextFactory();
+    HttpClient httpClient = new HttpClient(sslContextFactory);
+    httpClient.setConnectTimeout(credentials.getConnectTimeout());
+    if (!Strings.isNullOrEmpty(credentials.getProxyUrl())) {
+      Authenticator.setProxy(credentials, httpClient);
+    }
+
+    try {
+      httpClient.start();
+      Request request = httpClient.newRequest(explainUrl)
+          .method(HttpMethod.GET)
+          .header(
+              HttpHeader.AUTHORIZATION,
+              "Bearer " + oAuthInfo.getAccessToken())
+          .header(
+              HttpHeader.CONTENT_TYPE,
+              "application/json");
+      ContentResponse response = request.send();
+      String responseContent = response.getContentAsString();
+      if (response.getStatus() != HttpURLConnection.HTTP_OK) {
+        throw new IOException(String.format(
+            "Failed to get Query Plan. Status='%s', Response='%s'",
+            response.getStatus(), responseContent));
+      }
+      return new Gson().fromJson(responseContent, QueryPlanResponse.class);
+    } finally {
+      httpClient.stop();
+    }
+  }
+
+  /**
+   * Holds query plan response from Salesforce REST API.
+   */
+  public static class QueryPlanResponse {
+    private List<QueryPlan> plans;
+
+    public List<QueryPlan> getPlans() {
+      return plans;
+    }
+  }
+
+  /**
+   * Holds single query plan details.
+   */
+  public static class QueryPlan {
+    private double relativeCost;
+    private long cardinality;
+    private String leadingOperationType;
+
+    public double getRelativeCost() {
+      return relativeCost;
+    }
+
+    public long getCardinality() {
+      return cardinality;
+    }
+
+    public String getLeadingOperationType() {
+      return leadingOperationType;
+    }
   }
 }
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
@@ -48,6 +48,8 @@ import io.cdap.plugin.salesforce.plugin.OAuthInfo;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.BulkConnectionRetryWrapper;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSourceConstants;
 import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSplitUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -71,6 +73,7 @@ public class SalesforceBatchSource extends
   public static final String NAME = "Salesforce";
 
   private final SalesforceSourceConfig config;
+  private static final Logger LOG = LoggerFactory.getLogger(SalesforceBatchSource.class);
   private Schema schema;
   private MapToRecordTransformer transformer;
   private Set<String> jobIds = new HashSet<>();
@@ -139,7 +142,7 @@ public class SalesforceBatchSource extends
 
     authenticatorCredentials = config.getConnection().getAuthenticatorCredentials();
     List<SalesforceSplit> querySplits =
-        getSplits(config, authenticatorCredentials, context.getLogicalStartTime(), oAuthInfo);
+        getSplits(config, authenticatorCredentials, context.getLogicalStartTime(), oAuthInfo, false);
     querySplits.stream().forEach(salesforceSplit -> jobIds.add(salesforceSplit.getJobId()));
     context.setInput(Input.of(config.getReferenceNameOrNormalizedFQN(orgId, sObjectName),
         new SalesforceInputFormatProvider(
@@ -148,10 +151,22 @@ public class SalesforceBatchSource extends
 
   public static List<SalesforceSplit> getSplits(
       SalesforceSourceConfig config, AuthenticatorCredentials authenticatorCredentials,
-      long logicStartTime, OAuthInfo oAuthInfo) {
+      long logicStartTime, OAuthInfo oAuthInfo, boolean pkChunkCountCheck) {
     String query = config.getQuery(logicStartTime, oAuthInfo);
     BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+
     boolean enablePKChunk = config.getEnablePKChunk();
+
+    if (enablePKChunk && pkChunkCountCheck) {
+      enablePKChunk = SalesforceSplitUtil.hasRequiredCountForPkChunking(
+          query, authenticatorCredentials, config.getChunkSize());
+      if (!enablePKChunk) {
+        LOG.info(
+            "PK chunking skipped: record count is below threshold {} ",
+                config.getChunkSize());
+      }
+    }
+
     if (enablePKChunk) {
       String parent = config.getParent();
       int chunkSize = config.getChunkSize();
--- a/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
+++ b/src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
@@ -26,6 +26,8 @@ import com.sforce.async.ContentType;
 import com.sforce.async.JobInfo;
 import com.sforce.async.JobStateEnum;
 import com.sforce.async.OperationEnum;
+import com.sforce.soap.partner.PartnerConnection;
+import com.sforce.soap.partner.QueryResult;
 import dev.failsafe.FailsafeException;
 import dev.failsafe.RetryPolicy;
 import dev.failsafe.TimeoutExceededException;
@@ -33,9 +35,13 @@ import io.cdap.plugin.salesforce.BulkAPIBatchException;
 import io.cdap.plugin.salesforce.InvalidConfigException;
 import io.cdap.plugin.salesforce.SObjectDescriptor;
 import io.cdap.plugin.salesforce.SalesforceBulkUtil;
+import io.cdap.plugin.salesforce.SalesforceConnectionUtil;
+import io.cdap.plugin.salesforce.SalesforceConstants;
 import io.cdap.plugin.salesforce.SalesforceQueryUtil;
 import io.cdap.plugin.salesforce.authenticator.Authenticator;
 import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
+import io.cdap.plugin.salesforce.parser.SalesforceQueryParser;
+import io.cdap.plugin.salesforce.plugin.OAuthInfo;
 import io.cdap.plugin.salesforce.plugin.source.batch.SalesforceSplit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -263,8 +269,76 @@ public final class SalesforceSplitUtil {
     return false;
   }
 
-  // This is added only for UCS use case.
-  private static boolean isCustomObject(String sobjectName) {
+  /**
+   * Determines whether PK chunking should be enabled or disabled based on the
+   * estimated record count, object support, and query compatibility.
+   *
+   * @param query       the SOQL query
+   * @param credentials authenticator credentials for SOAP API
+   * @param threshold   the record count threshold for enabling PK chunking
+   * @return true if PK chunking should be enabled, false otherwise
+   */
+  public static boolean hasRequiredCountForPkChunking(
+      final String query,
+      final AuthenticatorCredentials credentials,
+      final long threshold) {
+    try {
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+      String countQuery = SalesforceQueryUtil.createCountQuery(query);
+
+      try {
+        SalesforceQueryUtil.QueryPlanResponse planResponse =
+            SalesforceQueryUtil.getQueryPlan(countQuery, credentials);
+        if (planResponse != null && planResponse.getPlans() != null
+            && !planResponse.getPlans().isEmpty()) {
+          SalesforceQueryUtil.QueryPlan leadingPlan =
+              planResponse.getPlans().get(0);
+          LOG.debug(
+              "PK Chunking Query Plan: leading operation is '{}', "
+                  + "cardinality is {}, cost is {}",
+              leadingPlan.getLeadingOperationType(),
+              leadingPlan.getCardinality(),
+              leadingPlan.getRelativeCost());
+          if (leadingPlan.getCardinality() >= threshold
+              || leadingPlan.getRelativeCost() >= 1.0) {
+            LOG.info(
+                "PK Chunking: Query Plan indicates high volume or "
+                    + "non-selective scan. Auto-enabling PK Chunking.");
+            return true;
+          }
+        }
+      } catch (Exception e) {
+        LOG.warn(
+            "PK Chunking: Query Plan check failed, defaulting to true.",
+            e);
+        return true;
+      }
+
+      PartnerConnection partnerConnection =
+          SalesforceConnectionUtil.getPartnerConnection(credentials);
+      QueryResult result = partnerConnection.query(countQuery);
+      int recordCount = result.getSize();
+      LOG.debug(
+          "PK Chunking validation: object '{}' has {} records, "
+              + "threshold is {}",
+          sObjectName, recordCount, threshold);
+      return recordCount >= threshold;
+    } catch (Exception e) {
+      LOG.warn(
+          "PK Chunking validation: unexpected error during COUNT() query "
+              + "check, falling back to default PK chunking",
+          e);
+      return true;
+    }
+  }
+
+  /**
+   * Helper method to check if sobject is custom object.
+   *
+   * @param sobjectName name of the sobject
+   * @return true if custom, false otherwise
+   */
+  private static boolean isCustomObject(final String sobjectName) {
     return sobjectName.toLowerCase().endsWith("__c");
   }
 }
--- a/src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
+++ b/src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
@@ -16,9 +16,25 @@
 package io.cdap.plugin.salesforce;
 
 import com.google.common.collect.ImmutableMap;
+import io.cdap.plugin.salesforce.authenticator.Authenticator;
+import io.cdap.plugin.salesforce.authenticator.AuthenticatorCredentials;
+import io.cdap.plugin.salesforce.plugin.OAuthInfo;
+import io.cdap.plugin.salesforce.plugin.source.batch.util.SalesforceSplitUtil;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpMethod;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
 
+import java.net.HttpURLConnection;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -32,6 +48,15 @@ import java.util.stream.IntStream;
 /**
  * Tests for {@link SalesforceQueryUtil}.
  */
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({
+    SalesforceConnectionUtil.class,
+    Authenticator.class,
+    SalesforceQueryUtil.class,
+    OAuthInfo.class,
+    HttpClient.class,
+    org.eclipse.jetty.util.component.AbstractLifeCycle.class
+})
 public class SalesforceQueryUtilTest {
 
   @Test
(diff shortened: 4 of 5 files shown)


src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,43 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+      }
+      String countQuery = SalesforceQueryUtil.createCountQuery(query);
+      PartnerConnection partnerConnection = SalesforceConnectionUtil.getPartnerConnection(credentials);
+      QueryResult result = partnerConnection.query(countQuery);
+      int recordCount = result.getSize();
+      LOG.debug("PK Chunking auto-decision: object '{}' has {} records, threshold is {}",
+              sObjectName, recordCount, threshold);
+      return recordCount >= threshold;
+    } catch (ConnectionException e) {
+      LOG.warn("PK Chunking auto-decision: failed to execute count query, falling back to no PK chunking", e);
```
vikasrathee-cs: falling back to default PK chunking setting, default should be true

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -148,10 +151,22 @@ public void prepareRun(BatchSourceContext context) {
...
     BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+
     boolean enablePKChunk = config.getEnablePKChunk();
+
+    if (enablePKChunk && autoEnablePKChunk) {
+      enablePKChunk = SalesforceSplitUtil.shouldAutoPKChunk(
+          query, authenticatorCredentials, SalesforceSourceConstants.AUTO_PK_CHUNK_THRESHOLD);
+      if (!enablePKChunk) {
+        LOG.info("PK chunking skipped: record count is below threshold {} or "
+                + "object/query does not support it.",
```
vikasrathee-cs: remove object does not support it as this will be called only for supported objects

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,138 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
     Assert.assertEquals("record3", results.get(2).get("Name"));
     Assert.assertEquals("record4", results.get(3).get("Name"));
   }
+
+  /**
+   * Test: UI toggle TRUE + isPKChunkable=true + records > AUTO_PK_CHUNK_THRESHOLD
```
vikasrathee-cs: isPKChunkable, fix the name with actual name

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,43 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+    } catch (ConnectionException e) {
+      LOG.warn("PK Chunking auto-decision: unexpected error, falling back to default PK chunking", e);
+      return false;
+    } catch (Exception e) {
+      LOG.warn("PK Chunking auto-decision: unexpected error, falling back to default PK chunking", e);
+      return false;
+    }
+  }
+
+
```
vikasrathee-cs: remove extra space

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -148,10 +151,21 @@ public void prepareRun(BatchSourceContext context) {
...
     String query = config.getQuery(logicStartTime, oAuthInfo);
     BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+
     boolean enablePKChunk = config.getEnablePKChunk();
+
+    if (enablePKChunk && autoEnablePKChunk) {
+      enablePKChunk = SalesforceSplitUtil.shouldAutoPKChunk(
+          query, authenticatorCredentials, SalesforceSourceConstants.AUTO_PK_CHUNK_THRESHOLD);
+      if (!enablePKChunk) {
+        LOG.info("PK chunking skipped: record count is below threshold {} or ",
```
vikasrathee-cs: remove or

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,153 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
...
+   * Test: UI toggle TRUE + autoEnablePKChunk=true + records > AUTO_PK_CHUNK_THRESHOLD
+   * → count check runs → PK Chunking enabled
+   *
+   * NOTE: Since we cannot create 10M records in an integration test, this test
+   * verifies the pipeline runs correctly when PK chunking is enabled via config.
+   * The threshold logic is covered by unit tests in SalesforceSplitUtilTest.
+   */
+  @Test
+  public void pkChunking_aboveThreshold_returnsAll() throws Exception {
+    // Arrange
```
vikasrathee-cs: remove these comments //asser, //arrange, just have a blank line in between to segregate these

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,42 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+  /**
+   * Determines whether PK chunking should be enabled automatically based on record count,
+   * object support, and query compatibility.
+   *
+   * @param query       the SOQL query
+   * @param credentials authenticator credentials for SOAP API
+   * @param threshold   the record count threshold above which PK chunking is enabled
+   * @return true if PK chunking should be auto-enabled, false otherwise
+   */
+  public static boolean shouldAutoPKChunk(String query, AuthenticatorCredentials credentials, long threshold) {
```
vikasrathee-cs: shouldAutoDetectPKChunk, use this name, this is more apt for this. Also change the variable name used later in the PR for its result.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,42 @@ public static boolean isPkChunkingSupported(String sobjectName) {
     return false;
   }
 
+  /**
+   * Determines whether PK chunking should be enabled automatically based on record count,
```
vikasrathee-cs: this should be enable/disable

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,42 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+   *
+   * @param query       the SOQL query
+   * @param credentials authenticator credentials for SOAP API
+   * @param threshold   the record count threshold above which PK chunking is enabled
+   * @return true if PK chunking should be auto-enabled, false otherwise
+   */
+  public static boolean shouldAutoDetectPKChunk(String query, AuthenticatorCredentials credentials, long threshold) {
+    try {
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+      if (!isPkChunkingSupported(sObjectName)) {
```
vikasrathee-cs: sobjectName.toLowecase

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,42 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+      int recordCount = result.getSize();
+      LOG.debug("PK Chunking auto-decision: object '{}' has {} records, threshold is {}",
+              sObjectName, recordCount, threshold);
+      return recordCount >= threshold;
+    } catch (ConnectionException e) {
+      LOG.warn("PK Chunking auto-decision: unexpected error, falling back to default PK chunking", e);
+      return true;
+    } catch (Exception e) {
+      LOG.warn("PK Chunking auto-decision: unexpected error, falling back to default PK chunking", e);
+      return false;
```
vikasrathee-cs: here also it should be true

vikasrathee-cs (approved):

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
```diff
@@ -55,6 +55,7 @@ public class SalesforceSourceConstants {
   public static final int MAX_PK_CHUNK_SIZE = 250000;
   public static final int DEFAULT_PK_CHUNK_SIZE = 100000;
   public static final int MIN_PK_CHUNK_SIZE = 1;
+  public static final long AUTO_PK_CHUNK_THRESHOLD = 1_000_000;
```
Sunish-Dahiya: 1M is too high, we should use the chunk size(that we generally use to create chunks) as a threshold here

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,39 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+   *
+   * @param query       the SOQL query
+   * @param credentials authenticator credentials for SOAP API
+   * @param threshold   the record count threshold above which PK chunking is enabled
+   * @return true if PK chunking should be auto-enabled, false otherwise
+   */
+  public static boolean shouldAutoDetectPKChunk(String query, AuthenticatorCredentials credentials, long threshold) {
+    try {
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+      if (!isPkChunkingSupported(sObjectName.toLowerCase())) {
```
Sunish-Dahiya: won't this check be already there in the code, I am assuming shouldAutoDetectPKChunk would only be called when all the conditions of chunking has already been satisfied

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,39 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+        LOG.debug("PK Chunking auto-decision: object '{}' is not supported for PK chunking", sObjectName);
+        return false;
+      }
+      if (SalesforceQueryParser.isRestrictedPKQuery(query)) {
```
Sunish-Dahiya: same as above

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,39 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+  /**
+   * Determines whether PK chunking should be enable/disable automatically based on record count,
+   * object support, and query compatibility.
+   *
+   * @param query       the SOQL query
+   * @param credentials authenticator credentials for SOAP API
+   * @param threshold   the record count threshold above which PK chunking is enabled
+   * @return true if PK chunking should be auto-enabled, false otherwise
+   */
+  public static boolean shouldAutoDetectPKChunk(String query, AuthenticatorCredentials credentials, long threshold) {
```
Sunish-Dahiya: We are not auto detecting chunking, instead it's a validation check for PK chunking.
We can use more intuitive names like `hasRequiredCountForPkChunking`

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,6 +268,39 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+    try {
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+      if (!isPkChunkingSupported(sObjectName.toLowerCase())) {
+        LOG.debug("PK Chunking auto-decision: object '{}' is not supported for PK chunking", sObjectName);
+        return false;
+      }
+      if (SalesforceQueryParser.isRestrictedPKQuery(query)) {
+        LOG.debug("PK Chunking auto-decision: query contains restricted clauses, skipping PK chunking");
```
Sunish-Dahiya: record count criteria not auto decision, apply everywhere

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
```diff
@@ -227,4 +227,18 @@ public void testCreateSObjectIdQuery() {
 
     Assert.assertEquals("SELECT Id " + fromClause, sObjectIdQuery);
   }
+
+  @Test
+  public void testCreateCountQuery() {
```
Sunish-Dahiya: can we please follow same naming convention across all tests?

Test names should summarize the behavior being tested and its expected outcome.

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
```diff
@@ -227,4 +227,18 @@ public void testCreateSObjectIdQuery() {
...
+    String query = "SELECT Id,Name,SomeField FROM sObjectName WHERE LastModifiedDate>=2019-04-12T23:23:23Z";
+    String countQuery = SalesforceQueryUtil.createCountQuery(query);
```
Sunish-Dahiya: Structure test in arrange, act, assert blocks , you can use empty line as separatot between these blocks, Comment applies to all

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,135 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
...
+  /**
+   * Test: Custom object (__c suffix) with UI toggle TRUE + records below threshold
+   * → count check runs → PK Chunking skipped (below threshold)
+   * → records returned correctly via normal batch processing
+   *
+   * Custom objects are supported for PK chunking (isPkChunkingSupported returns true
+   * for __c suffix), but threshold check still applies.
+   */
+  @Test
+  public void pkChunking_customObject_returnsAll() throws Exception {
```
Sunish-Dahiya: Please follow naming and structure guideline as described in below comments

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
```diff
@@ -83,6 +83,18 @@ public static String createSObjectIdQuery(String query) {
...
   }
 
+  /**
+   * Creates a COUNT query from an existing SOQL query.
+   * Replaces the SELECT fields with COUNT() while preserving the FROM and WHERE clauses.
+   *
+   * @param query the original SOQL query
+   * @return a COUNT SOQL query string
+   */
+  public static String createCountQuery(String query) {
```
Sunish-Dahiya, line 107: How long does this query takes when we use filters?

Can you add tests cases with before and after time with high record count
1. Test table having 30-40M records and with a filter query
2. Table with 10M records with a filter query.

Basically the things that we want to tests is how does the query cost scales with record count.

Sunish-Dahiya (changes requested):

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
harishhk107: applied `testMethodName_stateUnderTest_expectedBehavior` pattern 
Added blank lines for Arrange/Act/Assert

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
harishhk107: sure

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
harishhk107: sure

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSourceConstants.java
harishhk107: While we could lower the threshold to the chunk size, we have tested the 1M threshold and found that it doesn't make a significant difference in execution time

Before ~13m
After ~13m

Sunish-Dahiya: 1M is too high from readstream perspective, for some cases we may not be able to read that much data in a single stream.

Also how are you measuring execution time? in DTS it would definitely change as with chunking we will process 1M record in parallel and without chunking it would happen in a sequential manner

harishhk107: I’m currently testing this locally on a CDAP pipeline.

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
harishhk107: Reason for `Limit`: The primary reason for adding the `LIMIT` clause is Salesforce’s hard-enforced 120-second synchronous query execution timeout. Without it, standard `COUNT()` queries on large tables often exceed this limit and raise a `QUERY_TIMEOUT` exception, causing pipeline failures.

Behavior After Code Changes: Because the query stops scanning once it hits the threshold, running this on a table with 10M records versus a table with 40M records will yield similar performance (assuming both have at least 100k records). The database will not perform a full scan of the 10M or 40M records. Therefore, performance tests will only show scaling behavior up to the threshold limit, regardless of the total table volume or the filters used.

Observations: In our observations, these count queries with the `LIMIT` clause typically take between 700ms and 2 seconds to complete depending on the table size, safely avoiding the 120-second synchronous timeout.

Sunish-Dahiya: LIMIT caps the maximum number of matching results returned, not the number of rows scanned to find those matches.
So I am not able to understand how adding limit will optimize the query.

harishhk107: if we were dealing with standard ANSI SQL databases (like PostgreSQL or MySQL). In standard SQL engines, SELECT COUNT(*) FROM table LIMIT X computes the full aggregation first by scanning all matching rows to produce a single scalar value, and then applies the LIMIT clause to that final single row. In that context, LIMIT does not optimize or reduce the number of rows scanned at all.

However, Salesforce’s SOQL query engine handles COUNT() natively in a different way:
When a LIMIT clause is appended to a COUNT() query in SOQL (e.g., SELECT COUNT() FROM Object WHERE ... LIMIT 100000), it acts as a short-circuit instruction for the Salesforce query optimizer. The Salesforce database engine stops scanning and counting records as soon as it reaches 100,000 matches that satisfy the filter criteria. It then immediately returns the QueryResult with size capped at 100,000.

Conclusion: If LIMIT operated like standard SQL (evaluating the full count first and capping the returned result set afterward), the database engine would still have to perform the full table scan of all 30+ million records. If it performed a full scan, the query would still take >120 seconds and trigger the exact same QUERY_TIMEOUT exception.

The fact that execution drops from >120 seconds (Timeout) down to <2 seconds is definitive proof that the Salesforce query engine stops scanning rows the moment the LIMIT threshold is reached.

> **Official Documentation References:**
> * For how `COUNT()` populates the `size` property directly: [SOQL COUNT() Documentation](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_count.htm)
> * For capping execution evaluations: [SOQL LIMIT Clause Documentation](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_limit.htm)

Sunish-Dahiya: It says the same thing as i mentioned above - `as soon as it reaches 100,000 matches that satisfy the filter criteria`

If a table has 1 million records and only 1 lakh are matching the filter criteria, to find out those 1lakh it may need to scan the whole table. So the query complexity will depend on the nature of filter and data, adding limit doesn't result in any optimisation in such case.
It can help in cases where lets say filter criteria return 1 million records but since we want to only check if it has atleast 1 lakh records then in those cases it would return early, as soon as it find 1 lakh record but finding 1 lakh record doesn't mean it won't do table scan.

Also the official refs attached in your comment do not mention anything related to query optimisation.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,8 +279,186 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+
+  /**
+   * Fetches query plan from Salesforce query explain REST API.
+   *
+   * @param query       SOQL query
+   * @param credentials credentials to authenticate the call
+   * @return query plan details
+   * @throws Exception in case of network or API failures
+   */
+  private static QueryPlanResponse getQueryPlan(
```
Sunish-Dahiya: This logic should be in query util,  not split util

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,8 +279,186 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+     * Estimated record cardinality.
+     */
+    private long cardinality;
+    /**
+     * Leading operation type (e.g., TableScan, Index).
+     */
+    private String leadingOperationType;
+
+    /**
+     * Getter for relativeCost.
```
Sunish-Dahiya: can we add something meaningful in the comment, or if it's straightforward then better to skip the javadoc for such methods.

This comment applies to all methods in this file

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,114 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
...
+            .put("Name", "record2")
+            .build())
+        .build();
+ 
+    addSObjects(sObjects, true);
+ 
+    ImmutableMap.Builder<String, String> properties = getBaseProperties("SalesforceReaderPKChunkBelowThreshold");
+    properties.put("enablePKChunk", "true");
+    properties.put(SalesforceSourceConstants.PROPERTY_QUERY, "SELECT Name FROM " + sObjectName);
+
```
Sunish-Dahiya: can you please follow best practices for writing and structuring UTs?
Applies to all

Sunish-Dahiya (changes requested):
On a high level change looks fine but can you please test all these scenario
1. When chunking is enabled for data less than <100K(old logic)
2. Chunking should be disabled for data less than <100K - compare time/performance of #1 and #2.
3. All other cases , adding few of them for reference: query is selective, query is expensive, query is selective but record count >100K

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
harishhk107: Sure

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
harishhk107: Okay

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
harishhk107: Sure

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,8 +270,74 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+          }
+        }
+      } catch (Exception e) {
+        LOG.warn(
+            "PK Chunking: Query Plan check failed, falling back to COUNT().",
+            e);
+      }
+
+      String countQuery = SalesforceQueryUtil.createCountQuery(query)
+          + " LIMIT " + threshold;
```
Sunish-Dahiya: why do we need this LIMIT clause now?

Sunish-Dahiya (changes requested):
Comment regarding unit tests is still not addressed

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,8 +270,74 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+  public static boolean hasRequiredCountForPkChunking(
+      final String query,
+      final AuthenticatorCredentials credentials,
+      final long threshold) {
+    try {
+      String sObjectName = SObjectDescriptor.fromQuery(query).getName();
+
+      try {
+        SalesforceQueryUtil.QueryPlanResponse planResponse =
+            SalesforceQueryUtil.getQueryPlan(query, credentials);
```
Sunish-Dahiya: we need a count query here not the actual query right?

Sunish-Dahiya: looks like we are relying on queryplanner to convert it into count query, this is not ideal, getQueryPlan should not mutate query it should find cost/plan of whatever query is passed to this function

harishhk107: Refactored getQueryPlan to be a pure function that does not mutate the input, and updated SalesforceSplitUtil to explicitly generate and pass the count query.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
harishhk107: Removed the LIMIT clause from the physical count query completely

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
harishhk107: Cleaned up the spacing in the test cases to ensure exactly one blank line separates the Arrange, Act, and Assert sections

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -263,8 +270,75 @@ public static boolean isPkChunkingSupported(String sobjectName) {
...
+          if (leadingPlan.getCardinality() >= threshold
+              || leadingPlan.getRelativeCost() >= 1.0) {
+            LOG.info(
+                "PK Chunking: Query Plan indicates high volume or "
+                    + "non-selective scan. Auto-enabling PK Chunking.");
+            return true;
+          }
+        }
+      } catch (Exception e) {
+        LOG.warn(
```
Sunish-Dahiya, line 311: does it make sense to return true if for some reason query plan fails? this will protect against the cases where query does a full scan but plan failed for it.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
```diff
@@ -55,6 +61,7 @@
 public final class SalesforceSplitUtil {
 
   private static final Logger LOG = LoggerFactory.getLogger(SalesforceSplitUtil.class);
+  private static final int HTTP_OK = 200;
```
Sunish-Dahiya: unused?

src/main/java/io/cdap/plugin/salesforce/SalesforceQueryUtil.java
```diff
@@ -106,4 +131,85 @@ private static String generateSObjectFilter(SObjectFilterDescriptor filterDescri
...
+          .method(HttpMethod.GET)
+          .header(
+              HttpHeader.AUTHORIZATION,
+              "Bearer " + oAuthInfo.getAccessToken())
+          .header(
+              HttpHeader.CONTENT_TYPE,
+              "application/json");
+      ContentResponse response = request.send();
+      String responseContent = response.getContentAsString();
+      if (response.getStatus() != 200) {
```
Sunish-Dahiya: constant

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
```diff
@@ -32,6 +38,10 @@
 /**
  * Tests for {@link SalesforceQueryUtil}.
  */
+@RunWith(PowerMockRunner.class)
```
Sunish-Dahiya, line 51: we don't have test for getQueryPlan

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,108 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
...
+        properties.build(), SalesforceBatchSource.NAME, "SalesforceBatch");
+    results.sort(Comparator.comparing(record -> record.get("Name")));
+
+    Assert.assertEquals(2, results.size());
+    Assert.assertEquals("record1", results.get(0).get("Name"));
+    Assert.assertEquals("record2", results.get(1).get("Name"));
+  }
+
+  @Test
+  public void getSplits_pkChunkEnabledOnCustomObjectBelowThreshold_chunkingSkipped() throws Exception {
```
Sunish-Dahiya, line 704: we should have tests for scenarios
1. getQueryPlan failed - default to true
2. count query failed - default to true

Sunish-Dahiya (changes requested):
Please address the comments before merging.

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/util/SalesforceSplitUtil.java
harishhk107: If getQueryPlan fails, it now catches the exception, logs a warning, and immediately returns true (enabling PK chunking to prevent potential timeouts

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
harishhk107: added - getQueryPlan_success_returnsQueryPlanResponse

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
harishhk107: added
hasRequiredCountForPkChunking_queryPlanFails_defaultsToTrue
hasRequiredCountForPkChunking_countQueryFails_defaultsToTrue

harishhk107:
Add all the testcases in the SalesforceQueryUtilTest to reuse the existing PowerMockito class-level configuration for mocking query plan REST calls and

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
```diff
@@ -227,4 +252,109 @@ public void testCreateSObjectIdQuery() {
...
+  @Test
+  public void getQueryPlan_success_returnsQueryPlanResponse() throws Exception {
+    String query = "SELECT COUNT() FROM Account";
+    AuthenticatorCredentials credentials = Mockito.mock(AuthenticatorCredentials.class);
+    Mockito.when(credentials.getConnectTimeout()).thenReturn(3000);
+    OAuthInfo oAuthInfo = PowerMockito.mock(OAuthInfo.class);
+    Mockito.when(oAuthInfo.getInstanceURL()).thenReturn("https://instance.salesforce.com");
+    Mockito.when(oAuthInfo.getAccessToken()).thenReturn("test-token");
+    PowerMockito.mockStatic(Authenticator.class);
+    PowerMockito.when(Authenticator.getOAuthInfo(credentials)).thenReturn(oAuthInfo);
```
Sunish-Dahiya, line 297: still not following structuring guidelines

src/test/java/io/cdap/plugin/salesforce/SalesforceQueryUtilTest.java
```diff
@@ -227,4 +252,109 @@ public void testCreateSObjectIdQuery() {
...
+        .thenReturn("SELECT COUNT() FROM Opportunity");
+    SalesforceQueryUtil.QueryPlanResponse planResponse = Mockito.mock(SalesforceQueryUtil.QueryPlanResponse.class);
+    SalesforceQueryUtil.QueryPlan plan = Mockito.mock(SalesforceQueryUtil.QueryPlan.class);
+    Mockito.when(plan.getCardinality()).thenReturn(10L);
+    Mockito.when(plan.getRelativeCost()).thenReturn(0.1);
+    Mockito.when(planResponse.getPlans()).thenReturn(Collections.singletonList(plan));
+    PowerMockito.when(SalesforceQueryUtil.getQueryPlan(Mockito.anyString(), Mockito.any()))
+        .thenReturn(planResponse);
+    PowerMockito.mockStatic(SalesforceConnectionUtil.class);
+    PowerMockito.when(SalesforceConnectionUtil.getPartnerConnection(credentials))
```
Sunish-Dahiya, line 353: same as above

src/test/java/io/cdap/plugin/salesforce/etl/SalesforceBatchSourceETLTest.java
```diff
@@ -624,4 +624,108 @@ private void testPKChunk(ImmutableMap<String, String> properties) throws Excepti
...
+        properties.build(), SalesforceBatchSource.NAME, "SalesforceBatch");
+    results.sort(Comparator.comparing(record -> record.get("Name")));
+
+    Assert.assertEquals(2, results.size());
+    Assert.assertEquals("record1", results.get(0).get("Name"));
+    Assert.assertEquals("record2", results.get(1).get("Name"));
+  }
+
+  @Test
+  public void getSplits_pkChunkDisabledViaUI_chunkingNotApplied() throws Exception {
```
Sunish-Dahiya: doesn't matter via UI or some other client

harishhk107: remned to - getSplits_pkChunkDisabled_chunkingNotApplied

src/main/java/io/cdap/plugin/salesforce/plugin/source/batch/SalesforceBatchSource.java
```diff
@@ -148,10 +151,23 @@ public void prepareRun(BatchSourceContext context) {
...
+      long logicStartTime, OAuthInfo oAuthInfo, boolean pkChunkCountCheck) {
     String query = config.getQuery(logicStartTime, oAuthInfo);
     BulkConnection bulkConnection = SalesforceSplitUtil.getBulkConnection(authenticatorCredentials);
+
     boolean enablePKChunk = config.getEnablePKChunk();
+
+    if (enablePKChunk && pkChunkCountCheck) {
+      int chunkSize = config.getChunkSize();
+      enablePKChunk = SalesforceSplitUtil.hasRequiredCountForPkChunking(
+          query, authenticatorCredentials, chunkSize);
```
Sunish-Dahiya: this can be inlined, we don't need a new variable

Sunish-Dahiya (approved):

